PR #17, opened 2014-01-13
allow for customization of suspension policy

lucasmeijer:
requires latest Mono.Debugging.Soft


lucasmeijer:
closing for now, looks like we no longer need this.

PR #54, opened 2015-06-01
[SoftDebugger] Support reloading of .mdb files that are updated during debugging

lukaszunity:


jstedfast:
Hi Lukasz!

It seems like a better fix might be to simply keep a table of timestamps for the mdb files to use for the LastWriteTime check, then there would be no need to copy the files.

Or am I not understanding the patch quite right?

lukaszunity:
Hi Jeffrey

I see that my comments in the code do not fully explain the copying. The problem is on Windows some users are experiencing that they are unable to overwrite the existing .mdb because it is open. So the copy is made to make sure that the original file can be overwritten.

jstedfast:
Could the patch perhaps only copy the files to a temp location if the platform is Windows?

I'm assuming that this feature is to implement some sort of edit-and-continue type feature?

lukaszunity:
Yes, copying only when the platform is Windows make sense. This is for when scripts are edited in Unity while the debugger is attached, then Unity rebuilds and reloads the domain, but the mdb files are not reloaded and for some users copying the mdb files into the final directory also fails.

Should I submit a new PR with a Windows check?

jstedfast:
Yea, that would be awesome... thanks!

PR #55, opened 2015-06-03
[SoftDebugger] Support reloading of .mdb files that are updated during debugging v2

lukaszunity:
Only copy .mdb files on Windows. Use timestamps on other platforms.

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -55,6 +55,7 @@ namespace Mono.Debugging.Soft
 		readonly Dictionary<long,ObjectMirror> activeExceptionsByThread = new Dictionary<long, ObjectMirror> ();
 		readonly Dictionary<EventRequest, BreakInfo> breakpoints = new Dictionary<EventRequest, BreakInfo> ();
 		readonly Dictionary<string, MonoSymbolFile> symbolFiles = new Dictionary<string, MonoSymbolFile> ();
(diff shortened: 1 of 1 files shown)


DavidKarlas:
I would like to make one suggestion...
Currently CheckBetterMatch method is very inefficient(I never profiled code but feels just wrong) at end: https://github.com/mono/debugger-libs/blob/master/Mono.Debugging.Soft/SoftDebuggerSession.cs#L2548-L2579
It foreaches over all files(until correct one is founds) and ALL METHODS IN MDB and then do actual check if it's inside file we are interested in and do actually necessary check.

What I would like to suggest is to load all this data(Sources files names and methods informations/files/columns) into "internal format" of SoftDebuggerSession.cs.

What I have in mind is very simple...
```
class SourceFileMdbInfo
{
string FullFilePath;
byte[] Hash;
List<MethodMdbInfo> Methods;
}

class MethodMdbInfo
{
List<LineNumberEntry> SequencePoints;
}
Dictionary<string(fileNameNotFullPath, List<SourceFileMdbInfo>>;
```
see https://github.com/mono/debugger-libs/blob/master/Mono.Debugging.Soft/SoftDebuggerSession.cs#L2553-L2554 for why this dictionary is necessary(to support not full path equal if hash matches).

So once this data is loaded into this internal format... We can close mdb and keep checking just fileTimeStamp on all platforms including Windows.

lukaszunity:
I like your suggestion of loading the mdb's once better and I am going to implement this new approach instead.

Issue #57, opened 2015-06-22
Mismatch between domain reload handling in soft debugger client and agent.


lukaszunity:
I'm submitting this issue as a starting point for a discussion on how to fix a mismatch between how the soft debugger client and agent handle domain reloads.

The debugger agent registers new types (mirrors) with a type + domain pair.
https://github.com/mono/mono/blob/master/mono/mini/debugger-agent.c#L2276
and the types are marked as unloaded when a domain is unloaded.
https://github.com/mono/mono/blob/master/mono/mini/debugger-agent.c#L2210

In the client there is a single cache for all the types across domains
https://github.com/mono/debugger-libs/blob/master/Mono.Debugging.Soft/SoftDebuggerSession.cs#L60
and the types are removed when an assembly is unloaded.
https://github.com/mono/debugger-libs/blob/master/Mono.Debugging.Soft/SoftDebuggerSession.cs#L1879

This causes issues where the agent can unload a type associated with a domain, but if the assembly for the type is not unloaded (due to it being used in another domain), then the client will have a reference to a type that is marked as unloaded in the agent.

Zoltan Varga told me the following when I asked whether this was problem in the debugger agent: "Conceptually, assemblies are loaded into each domain they are referenced, so a reference to a type is really a reference to a type+domain pair, this is how the runtime code works. I think the problem is in the debugger-libs code, which should handle this the same way as the runtime code does, i.e. keep things in domain specific caches, and invalidate those caches when a domain is unloaded, not when an assembly is unloaded."

Would you accept PRs that address this issue and if so, do you have any suggestions for how to go about fixing it?

lukaszunity:
Any thoughts on this? I will start work on this soon in our own branch and would like to coordinate the effort on getting it into master.

DavidKarlas:
I didn't investigate much into this... But if we get all info needed over SDB protocol to implement this for example to which AppDomain Assembly Load/Unload belongs and to which AppDomain/Assembly TypeMirrors belong... I see no reason for not accepting this...

lukaszunity:
Sounds good. If possible, I will open smaller PRs with incremental changes to fix this issue.

lukaszunity:
I had a look into this issue and found out that in order to fix this issue, the debugger protocol has to be updated to include the domain id for a type loads and possibly for assemblies as well. Also, other classes that call into SoftDebuggerSession.GetType do not take the domain into account.

[...]

PR #60, opened 2015-09-03
[soft-debugger-session] Domain unload type fix.

lukaszunity:
Maintain list of loaded assemblies for each domain to remove unloaded types correctly on domain unload.

As discussed in issue https://github.com/mono/debugger-libs/issues/57

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -52,6 +52,7 @@ namespace Mono.Debugging.Soft
 	{
 		readonly Dictionary<Tuple<TypeMirror, string>, MethodMirror[]> overloadResolveCache = new Dictionary<Tuple<TypeMirror, string>, MethodMirror[]> ();
 		readonly Dictionary<string, List<TypeMirror>> source_to_type = new Dictionary<string, List<TypeMirror>> (PathComparer);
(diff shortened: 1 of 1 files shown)


lukaszunity:
I've updated this PR also remove types on Assembly Unload, to match the logic before opening this PR.

DavidKarlas:
Why is removal when Assembly Unload event happens needed? Won't UnloadDomain happen soon after?

Only thing that annoys me atm with this PR... is that when 10 assemblies are loaded at start(one after another) of program we will be executing extra SDB commands to fetch assemblies in domains 10 times... Maybe we could lazy fetch domain assemblies 1st time some type is loaded(if code gets too ugly... maybe we can live with existing solution in this PR)

I also noticed that types that need unloading are determined by assembly location...(this is existing code)
Doesn't this mean that if type System.Net.Socket was loaded in DomainRoot and is part of /path/to/System.dll will be unloaded when some other AppDomain is unloaded, which shouldn't be?

lukaszunity:
You are right that types might be incorrectly unloaded in the wrong domain and in fact, if the some other AppDomain contains all the assemblies that are in DomainRoot, then this PR is equivalent to doing types.Clear() HandleAppDomainUnloadEvents when unloading the other domain.

I don't think there is an good way to solve this issue domain unload issue without updating the debugger agent to send the domain ids for type loads.

I will fix this enum evalution issue in our own fork with a work-around and then revisit the issue with a the proper fix in the future.

PR #61, opened 2015-09-11
[soft-debugger-session] Check if hash is null, otherwise SequenceEqual will throw an exception

lukaszunity:


DavidKarlas:
Thank you!

PR #62, opened 2015-09-22
Clear breakpoint requests on assembly unload

lukaszunity:
The BreakpointEventRequests have a MethodMirror member and the method it represents is also unloaded when unloading the assembly and therefore no longer valid.

The BreakpointEventRequests are not used when resolving pending_bes in ResolveBreakpoints and a new request is added in InsertBreakpoint, so the breakpoint is correctly re-added when the assembly/type is loaded again.

Fixes ERR_UNLOADED errors when enabling/disabling breakpoints after a assembly unload.


lukaszunity:
I closed this PR because I determined that this didn't actually fix the ERR_UNLOADED issue. 

However, as it turns out this does actually fix the ERR_UNLOADED issue when Assembly Unload fix (PR https://github.com/mono/debugger-libs/pull/65) is applied.

DavidKarlas:
This also works without PR #65, right?

lukaszunity:
Yes, it works without PR #65, it just requires that HandleAssemblyUnload works correctly :smile:

PR #63, opened 2015-09-22
Set of bug fixed

nerzhulart:
The main points:
-  When evaluation of object throws an exception it's shown as child object of eval result (just meaningless string was shown before)
- AsyncOperationManager deadlock fixed. Is rare cases when exception is thrown durind operation the callback  is never invoked, so it caused hanging op this operation.
- Need to clear breakpoints before detach debugger session (required in CorDebug implementation, but I made it in common API)
  P.S. In some of last commits I moved back the CorDebug into main repo so you can skip the CorDebug files to review, they're already pushed into MonoDevelop repo preserving thier history.


Therzok:
Why was this not reviewed?

nerzhulart:
I'll make another PR with more clean commits (without formatting noise etc.)

PR #64, opened 2015-09-24
[soft-debugger-session] Support reloading of mdb files.

lukaszunity:
Solution follows the proposal here: https://github.com/mono/debugger-libs/pull/55

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -68,15 +68,14 @@ namespace Mono.Debugging.Soft
 			public LineNumberEntry[] SequencePoints { get; set; }
 		}
 
+		readonly Dictionary<string, Dictionary<string, List<MdbSourceFileInfo>>> mdbsSourceFileInfos = new Dictionary<string, Dictionary<string, List<MdbSourceFileInfo>>>();
 		readonly Dictionary<Tuple<TypeMirror, string>, MethodMirror[]> overloadResolveCache = new Dictionary<Tuple<TypeMirror, string>, MethodMirror[]> ();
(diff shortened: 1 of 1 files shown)


DavidKarlas:
I like where this is going but I would prefer if
```
readonly Dictionary<string, List<MdbSourceFileInfo>> fileToSourceFileInfos = new Dictionary<string, List<MdbSourceFileInfo>>();
readonly Dictionary<string, List<MdbSourceFileInfo>> mdbToSourceFileInfos = new Dictionary<string, List<MdbSourceFileInfo>>();
```
was replaced with
```
readonly Dictionary<string, Dictionary<string, List<MdbSourceFileInfo>>> mdbsSourceFileInfos = new Dictionary<string, Dictionary<string, List<MdbSourceFileInfo>>>();
```

Where 1st key would represent mdbFilePath.
2nd key would be same logic as now "fileName" and "fullpathFileName".

This would simplify UnloadMdbFile method.

But more importantly I have bad filling about linked files that are shared between projects/.dlls... So I would prefer if fileNames/MdbSourceFileInfo are stored per .mdb instead of somewhat shared between all .mdbs...

lukaszunity:
Thanks for your suggestions, I will take a look and either comment on this PR again or open a new PR.

DavidKarlas:
Also... Please maintain shape orientated programming :D
https://twitter.com/jbevain/status/536497814998908930

lukaszunity:
PR with your suggested changes while maintaining shape oriented programming.

DavidKarlas:
Thank you for this contribution!

jstedfast:
I'm so glad people are amused by my shape oriented programming :)

Therzok:
if shape is trapeze then I agree.

PR #65, opened 2015-09-24
Assembly unload fix

lukaszunity:
Store assembly locations on assembly load, as the assemblies have already been unloaded in the runtime when accessing them on assembly unload.

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -53,6 +53,7 @@ namespace Mono.Debugging.Soft
 		readonly Dictionary<Tuple<TypeMirror, string>, MethodMirror[]> overloadResolveCache = new Dictionary<Tuple<TypeMirror, string>, MethodMirror[]> ();
 		readonly Dictionary<string, List<TypeMirror>> source_to_type = new Dictionary<string, List<TypeMirror>> (PathComparer);
 		readonly Dictionary<long,ObjectMirror> activeExceptionsByThread = new Dictionary<long, ObjectMirror> ();
+		readonly Dictionary<AssemblyMirror, string> assemblyLocations = new Dictionary<AssemblyMirror, string>();
 		readonly Dictionary<EventRequest, BreakInfo> breakpoints = new Dictionary<EventRequest, BreakInfo> ();
 		readonly Dictionary<string, MonoSymbolFile> symbolFiles = new Dictionary<string, MonoSymbolFile> ();
 		readonly Dictionary<TypeMirror, string[]> type_to_source = new Dictionary<TypeMirror, string[]> ();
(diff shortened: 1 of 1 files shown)


DavidKarlas:
Are you sure this fixes anything? Because looking at https://github.com/mono/debugger-libs/blob/a129a42bcdd0f978fbc911b696d323f6911d600e/Mono.Debugging.Soft/SoftDebuggerSession.cs#L1842 and https://github.com/mono/debugger-libs/blob/a129a42bcdd0f978fbc911b696d323f6911d600e/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs#L25-L27.

Seems to me like like Location is loaded and cached at HandleAssemblyLoadEvents.
If this cache is not working because objects are not same... then also Dictionary lookup won't... So I'm not sure what is happening here :/

lukaszunity:
The cached AssemblyMirrors are invalided in VirtualMachine on AssemblyLoad/AssemblyUnload, which also means that the Location is no longer cached if it is not accessed before AssemblyUnload.

https://github.com/mono/debugger-libs/blob/master/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs#L731

When unloading an Assembly in the runtime we immediately notify the debugger agent using a profiler event.

https://github.com/mono/mono/blob/master/mono/metadata/assembly.c#L3157

The debugger agent the sends to the AssemblyUnload event to the SoftDebuggerSession, but while this happens the runtime continues running and will free the name for assembly image, possibly before we get the assembly unload event in the client.

https://github.com/mono/mono/blob/master/mono/metadata/image.c#L1645

Note: assembly->image is not set to null after releasing the memory associated with the image.

So when querying Assembly.Location in SoftDebuggerSession, the Assembly might not have been unloaded in the debugger agent types cache yet and since the assembly->image is not null, we end up accessing a garbage assembly->image->name pointer in mono_image_get_filename

https://github.com/mono/mono/blob/master/mono/mini/debugger-agent.c#L7596

On OSX I'm seeing complete garbage strings for Assembly.Location being sent over while debugging this issue and the dictionary look-ups then fail and things just break :-)

Side-note: We've have had this fix in our own branch of MD since 2012.

DavidKarlas:
I understand that calling [vm.conn.Assembly_GetLocation](https://github.com/mono/debugger-libs/blob/master/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs#L26) in UnloadEvent will return invalid value(since it's probably already destroyed by runtime)... But invalidating cached version of assemblies of domains is not affecting https://github.com/mono/debugger-libs/blob/master/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs#L436 which is used to store AssemblyMirror objects that has .Location value cached, hence it's not needed to call [vm.conn.Assembly_GetLocation](https://github.com/mono/debugger-libs/blob/master/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs#L26) at UnloadEvent...

I'm seeing valid Assembly.Location values here, so I'm not sure what bug are you referring to(about string for Assembly.Location being garbage)... Maybe that is reason for your problems...

lukaszunity:
Looks like you are right and I have gotten things a little mixed up. While I was fixing this I was also fixing an issue in our older debugger-agent which is already fixed in the latest version.

https://github.com/mono/mono/commit/f29c547b659c03126989622302dd664e0a29983f

So in our debugger agent we would send a wrong type id for the assembly unload to the client (which wasn't cached) and then we would get the issue AssemblyMirror.Location being garbage.

Closing this PR, as this fix is unnecessary.

PR #69, opened 2015-12-02
Modified using Mono.Cecil, ICSharpCode.NRefactory and NUnit from nuget

techcap:


DavidKarlas:
<Private>False</Private> were changed into <Private>True</Private>, which means copy files and that is not OK, because MonoDevelop or other projects might have custom distributions with additional fixes of Cecil and other stuff...
Unfortunately it's job of project using debugger-libs to copy those files...

Also can you please check this works fine with compiling MonoDevelop?

techcap:
I changed to use the Cecil in nuget package. So there's no dependency with other Cecil project, but only related to specific Cecil version which is declared in packages.config.
Do you want to have dependency of prereleased Cecil version?

DavidKarlas:
We can't load multiple Cecils, NRefactory... dlls... Ones loaded by MonoDevelop has to be used... API is probably same between MonoDevelop version of dlls and NuGet... but again... debugger-libs can't copy this dlls with them, aka no `<Private>True</Private>`

DavidKarlas:
Also, sorry to be like this, but before I even start considering/testing this PR for acceptance... I need you to test if this stuff work with MonoDevelop...

joj:
Using Mono.Cecil as a submodule is a hard requirement for us. There are several VS extensions that use different versions of cecil, so we change the assembly name on compile time to avoid collisions. Having cecil be used as a nuget will break that.

techcap:
How about ICsharpCode.NRefactory and nunit?
Now I have to add nunit in GAC and download ICsharpCode source code.
Is it also hard requirement?

joj:
We don't ship nunit and we haven't had that issue with nrefactory. We have had that issue with Cecil, though. It seems to be more common than the others.

techcap:
debugger-libs projects uses Mono.Cecil.dll without renaming and uses master branch of Mono.Cecil without touch. But one possible reason of using with source code would be the case of delayed deploy in nuget.
Mono.Cecil 0.9.6.1 released on April 30 in nuget. But the latest checkin is about a month ago. So if you had to use Mono.Cecil with source code, it would be that reason.
Thank you for kind explanation for newbie.

I'll pull request again without Mono.Cecil and ICsharpCOde.NRefactory.

PR #70, opened 2015-12-03
Fix nunit reference error

techcap:


jstedfast:
Cleaning up obsolete PR's.

PR #76, opened 2016-06-16
Updating from master, to be used in C7

joj:
Don't merge yet. Need some clarification on when and where this should ship


slluis:
Why are we merging this to cycle6 instead of just using whatever was referenced by cycle7?

joj:
This will break XS. It should only be used in VS. I had a patch in C6 that I need to move over to C7 (or C8, not sure) but that is based on master. This is only updating that old C6 branch to the latest in master.

joj:
Closing, reissuing against cycle7VS

PR #82, opened 2016-10-27
[Session] Fix every source file having a copy of every method instance.

Therzok:
This decreases memory usage by having only methods in a source file being used inside.

Based on my analysis of mdb writers, the mapping is one CompilationUnit
for each SourceFileEntry.

See MdbWriter and SymbolWriterImpl and grep for DefineDocument.

Bug 45792 - Exponential memory usage when loading mdb files


Therzok:
We could possibly patch up Cecil to offer methods per compilation unit, as 4.8 million iterations _could_ be costly. :)

PR #84, opened 2016-11-09
[SoftDebugger] Refactoring of methods invocation on mirrors

nerzhulart:
[SoftDebugger] New IInvocableMethodOwnerMirror interface to avoid duplicating of Invoke*, BeginInvoke*, EndInvoke* methods in invocable mirrors. All the API is now static methods taking IInvocableMethodOwnerMirror as the 1st parameter.  This refactoring allows to avoid multiple 'if (obj is *) else' in SoftDebuggerAdaptor.


nerzhulart:
Merge please this commit and then I'm going to extract all the static methods in ObjectMirror like Invoke*, BeginInvoke*, EndInvoke* into separate static class and make them to be extensions to IInvocableMethodOwnerMirror. It'll allow to save API compatibility with the old one.
Globally I had to make these refactoring because I started to rework AsyncOperationManager internals to .net Tasks (will PR this as well some later)

nerzhulart:
@Therzok Thanks for so quick reaction :) 
I've added two more commits.

DavidKarlas:
Could you please open pull request on files in `Mono.Debugger.Soft` subfolder(so all files except SoftDebuggerAdaptor.cs) against https://github.com/mono/mono? Which are in https://github.com/mono/mono/tree/master/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft

Unfortunately for historical reasons we are just copying those files here...

Mono.Debugger.Soft/Mono.Debugger.Soft/ObjectMirror.cs
```diff
@@ -142,37 +141,43 @@ public class ObjectMirror : Value {
...
+		// Public invocation API
+
+		public static Value InvokeMethod (IInvocableMethodOwnerMirror mirror, ThreadMirror thread, MethodMirror method, IList<Value> arguments, InvokeOptions options = InvokeOptions.None) {
+			return InvokeMethod (mirror, mirror.VirtualMachine, thread, method, mirror.GetThisObject (), arguments, options);
 		}
 
 		[Obsolete ("Use the overload without the 'vm' argument")]
-		public IAsyncResult BeginInvokeMethod (VirtualMachine vm, ThreadMirror thread, MethodMirror method, IList<Value> arguments, InvokeOptions options, AsyncCallback callback, object state) {
-			return BeginInvokeMethod (vm, thread, method, this, arguments, options, callback, state);
+		public static IAsyncResult BeginInvokeMethod (IInvocableMethodOwnerMirror mirror, VirtualMachine vm, ThreadMirror thread, MethodMirror method, IList<Value> arguments, InvokeOptions options, AsyncCallback callback, object state) {
```
Therzok: I'm not sure how I feel about the API break. Can we keep the old methods for backwards compat and mark them as obsolete?

nerzhulart: look at the next commit

nerzhulart: It's a reason. But do you support backward ABI compat between major releases of XS?
I can keep this ~10 methods as instance members but I guess it's really bad design, because if you need to implement invocations for one another type, you have to copy paste these 10 methods.

nerzhulart:
Hey, guys! Any news on this PR?

DavidKarlas:
I'm going to merge it... and hopefully won't run into problems later when I request to merge into Mono repo...

nerzhulart:
Thanks!

PR #85, opened 2016-12-20
Invocations refactoring

nerzhulart:
This refactoring originally was started to fix multiple deadlocks in AsyncOperationManager.
After some investigation I've decided to rewrite all of this code to .net Task API to avoid syncronization issues caused by bad desing and wrong using of ManualResetEvents everywhere.
The main changes:
* I'ive introduced a new EvaluatorExceptionThrownException that wraps target exception object and allows to inspect this object in watch pad when your eval has thrown an exception.
* AsyncOperationManager.Invoke is explicitly synchronous now and returns OperationResult<T>. Also it was synhronous before as well, but returned void and the call site have to obtain the result in other field and it was complicated.
* I've added a lot of logging into this code


nerzhulart:
Hey, guys! Any updates on this? Tests on windows go fine.

Mono.Debugging/Mono.Debugging.Evaluation/IAsyncOperation.cs
```diff
@@ -0,0 +1,22 @@
+namespace Mono.Debugging.Evaluation
+{
+	public interface IAsyncOperation
```
slluis: This seems to be unused.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationBase.cs
```diff
@@ -0,0 +1,88 @@
...
+		public OperationResult (TValue result, bool resultIsException)
+		{
+			Result = result;
+			ResultIsException = resultIsException;
+		}
+	}
+
+	public static class OperationResultEx
+	{
+		public static OperationResult<TValue> ThrowIfException<TValue> (this OperationResult<TValue> result, EvaluationContext ctx)
```
slluis: Why not just make it a member of OperationResult<TValue>?

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -2199,113 +2205,60 @@ public MethodCall (SoftEvaluationContext ctx, MethodMirror function, object obj,
...
-		
-		void EndInvoke ()
+		protected override Task<OperationResult<Value>> InvokeAsyncImpl (CancellationToken token)
 		{
 			try {
-				result = ((IInvocableMethodOwnerMirror) obj).EndInvokeMethodWithResult (handle);
-			} catch (InvocationException ex) {
-				if (!Aborting && ex.Exception != null) {
-					string ename = ctx.Adapter.GetValueTypeName (ctx, ex.Exception);
-					var vref = ctx.Adapter.GetMember (ctx, null, ex.Exception, "Message");
```
slluis: Why is this logic gone?

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
```diff
@@ -27,215 +27,172 @@
...
+
+
+		public event EventHandler<BusyStateEventArgs> BusyStateChanged = delegate {  };
+
+		void WaitAfterCancel (IAsyncOperationBase op)
 		{
-			Disposing = true;
-			lock (operationsToCancel) {
-				foreach (AsyncOperation op in operationsToCancel) {
-					op.InternalShutdown ();
```
slluis: All the Shutdown logic is gone, why?

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -2199,113 +2205,60 @@ public MethodCall (SoftEvaluationContext ctx, MethodMirror function, object obj,
...
-				exception = ex;
-			} catch (Exception ex) {
-				DebuggerLoggingService.LogError ("Error in soft debugger method call thread on " + GetInfo (), ex);
-				exception = ex;
-			} finally {
-				ctx.Session.StackVersion++;
-			}
-		}
-		
-		string GetInfo ()
```
slluis: More logic that has been removed for no apparent good reason.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncEvaluationTracker.cs
```diff
@@ -41,7 +41,7 @@ namespace Mono.Debugging.Evaluation
 	/// will then be made asynchronous and the Run method will immediately return an ObjectValue
 	/// with the Evaluating state.
 	/// </summary>
-	public class AsyncEvaluationTracker: RemoteFrameObject, IObjectValueUpdater, IDisposable
```
slluis: Why this change?

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationBase.cs
```diff
@@ -0,0 +1,88 @@
...
+			var exceptionTypeName = ctx.Adapter.GetValueTypeName (ctx, result.Result);
+			throw new EvaluatorExceptionThrownException (result.Result, exceptionTypeName);
+		}
+	}
+
+	public interface IAsyncOperationBase
+	{
+		Task RawTask { get; }
+		string Description { get; }
+		void AfterCancelled (int elapsedAfterCancelMs);
```
slluis: What is AfterCancelled for? it doesn't seem to have any implementation.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
nerzhulart: I've replaced simple Exception message to full exception object. So now you can walk through it, but before you can see only exception message

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
nerzhulart: If you mean logic of getting exception message so look at line 2230. SoftOperationResult wraps exception object to show it like ordinary object for user

slluis: I mean GetInfo(), which is used to log the method and type name when there is an invocation error.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncEvaluationTracker.cs
nerzhulart: There were no usages of AsyncEvaluationTracker by this interface.
image

nerzhulart: If you used by some dynamic way I can revert this change

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
slluis: How is that object shown to the user?

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationBase.cs
nerzhulart: I guess this approach makes the code more simple. OperationResult is just a data class without any internal logic

Mono.Debugging/Mono.Debugging.Evaluation/AsyncEvaluationTracker.cs
slluis: The important bit is that RemoteFrameObject is a MarshalByRefObject, and AsyncEvaluationTracker needs to be remotable.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
```diff
@@ -27,215 +27,172 @@
...
 
-			if (timeout > 0) {
-				if (!methodCall.WaitForCompleted (timeout)) {
-					bool wasAborted = methodCall.Aborted;
-					methodCall.InternalAbort ();
-					lock (operationsToCancel) {
-						operationsToCancel.Remove (methodCall);
-						ST.Monitor.PulseAll (operationsToCancel);
-					}
-					if (wasAborted)
```
nerzhulart: there is no Aborted property on AsyncOperation now, Cancellation is controlled by Task API

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
nerzhulart: Shutdown had nearly the same logic as the Cancel. I've unified them into Cancel.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
```diff
@@ -27,215 +27,172 @@
...
-			lock (operationsToCancel) {
-				foreach (AsyncOperation op in operationsToCancel) {
-					op.InternalShutdown ();
+			var desc = op.Description;
+			DebuggerLoggingService.LogMessage (string.Format ("Waiting for cancel of invoke {0}", desc));
+			try {
+				if (!op.RawTask.Wait (ShortCancelTimeout)) {
+					try {
+						BusyStateChanged (this, new BusyStateEventArgs {IsBusy = true, Description = desc});
+						op.RawTask.Wait (LongCancelTimeout);
```
nerzhulart: I've made a lot of experiments and found that many calls of Abort doesn't aborts evaluation if the first call didn't abort successfully. It's right for CorDebug and SBD as well. You may try to run eval of method with loop 
`void RunLoop() {
   int i = 0;
    while (true) {
      Console.WriteLine(i++);
   }
}
`
And try to abort it many times, you will fail with timeout always,

Mono.Debugging/Mono.Debugging.Evaluation/IAsyncOperation.cs
nerzhulart: That's right! Thanks

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
nerzhulart: It's an ObjectValue with its children that has IsException flag. It's shown in watchpad as ordinary object

Mono.Debugging/Mono.Debugging.Evaluation/AsyncEvaluationTracker.cs
nerzhulart: Can you please explain how is this working? where to look at the remoting?

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
nerzhulart: Seems that GetInfo used only for logging. Isnt't it enough to use Descriptions for this? Now I log all this states using this property

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationBase.cs
nerzhulart: It's used in CorDebug (monodevelop repo). It's done to ensure that we awaited enough amount of time after Abort() call. Otherwise we can get exceptions in CorDebug.

nerzhulart:
I have to say that this code works in Rider since November and the evaluation experience was greatly improved as well as performance. Before this we faced with deadlocks many times per day.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncEvaluationTracker.cs
slluis: The library is designed so that the debugging engine can run in a remote process, which communicates with the client using remoting. That's really up to the debugger engine implementation. In such scenario, the AsyncEvaluationTracker could be executing in the remote process and generating ObjectValue instances that are serialized into the client process and which may need to hold a reference to that remote object (https://github.com/mono/debugger-libs/blob/ccadee54b48bcf932b8d8e69176e2ee56f47886b/Mono.Debugging/Mono.Debugging.Evaluation/AsyncEvaluationTracker.cs#L85). This model was necessary with the old Mono debugger, but it is not with the soft debugger. However, the infrastructure is still there, in case a specific debugger implementation needs it.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationBase.cs
slluis: That looks unnecessary to me. In any case, up to you.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
slluis: I'm not sure if it is enough or not, it needs to be checked. I just don't want code to be removed if there isn't a good reason.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationBase.cs
nerzhulart: In this case ThrowIfException is more helper method rather than required to be in the class. It's IMHO

slluis:
Even if it works for Rider, I need to make sure that it doesn't cause any regression in XS. It is an important change in a critical piece of code, so we need to be very careful about it.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
nerzhulart: The main reasons to remove are simplifying and avoiding duplications.

nerzhulart:
Yes, of course! I just wanted to say that this code was not written a day ago and pushed without any testing

nerzhulart:
I've removed IAsyncOperation.cs and restore AsyncOperationTracker to be RemoteFrameObject.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
slluis: Nearly the same logic, but not the same. Let's say an invocation is made. While the caller is waiting for the invocation to end, the manager is disposed, so the call is canceled. But if the call can't really be cancelled, the caller will wait forever.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
slluis: Maybe in some cases will always timeout, maybe in others will not. In any case, we can't leave the "Busy" state until the debugger really stopped executing the method.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
slluis: Let's bring it back then.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
slluis: Ok

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
nerzhulart: Why do you think so? After Dispose is called the Invoke method on another thread will throw EvaluationAbortedException

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
nerzhulart: There is a big issue with your logic (that was used in MD now): if we try to abort some calculation that cannot be aborted gracefully (e.g. endless loop) the debugger may NOT to fire any signal to us - finished or aborted - and in this case we can wait forever. I've already faced with this that's why I've rewritten this to use timeouts. It's absolutely right for CorDebug, I'm not sure about SDB. Also ICorDebugEval has IsActive property, but after any times of Abort() on endless loop or some other long calculation it says us 'true'. I've read sources of .net core (because ICD impl in .net core is pretty the same as full framework) and they just wait 500 ms after Abort call and only after that your the ICD goes to consistent state. So waiting for IsActive == false may lead to forever awaiting.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
slluis: If the .Abort() call on the invocation fails, or can't be completed (it can happen, that's why there is all that abort retry and debugger busy mode infrastructure), then the invocation will never end, and EvaluationAbortedException will never be thrown.

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
slluis: >  if we try to abort some calculation that cannot be aborted gracefully (e.g. endless loop) the debugger may NOT to fire any signal to us - finished or aborted - and in this case we can wait forever.

That's is correct, and it is the whole point of having the debugger Busy mode. In SDB, if an invocation is in progress, we can't do anything else with the debugger. So, if the invocation can't be aborted, the only thing we can do is wait forever until the debuggee is able to end the invocation, or until the user stops the debugging session. When we enter the Busy mode, a dialog is shown to the user asking what to do (keep waiting for the call to end or stop the debugging session).

If this is not a problem for CorDebug, then you can add a flag to avoid waiting forever, but we need to keep the current behavior for SDB.

nerzhulart: I've restored the inifinite waiting. Now it should work as before. Also I've improved exception handling in SBD invocations. See commit "Invocation is awaited infinitely as it was before.  …"

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
nerzhulart: Done!

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
nerzhulart: Fixed. See below

nerzhulart:
Are there any new suggestions for this PR?

nerzhulart:
@slluis Hi! Any news on this?

slluis:
Sorry, it has been a busy week. I'll review it asap.

nerzhulart:
Ok! Thanks. Will wait.

nerzhulart:
Added a couple of commits. Now the Abort() is being called in loop as it was before.

nerzhulart:
Hi! Any news on this? 
I have new fixes for value modifications, but they are dependent on this

nerzhulart:
Added commits for Win32 debugger (moved from MD)

Mono.Debugging/Mono.Debugging.Evaluation/IAsyncOperation.cs
slluis: Still unused...

Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
```diff
@@ -27,215 +27,163 @@
...
-			Disposing = true;
-			lock (operationsToCancel) {
-				foreach (AsyncOperation op in operationsToCancel) {
-					op.InternalShutdown ();
+			var desc = op.Description;
+			DebuggerLoggingService.LogMessage (string.Format ("Waiting for cancel of invoke {0}", desc));
+			if (!op.RawTask.Wait (ShortCancelTimeout)) {
+				try {
+					ChangeBusyState (true, desc);
+					while (true) {
```
nerzhulart: >This loop should gracefully exit if AsyncOperationManager is disposed.
Done

nerzhulart: >Logic for switching to busy state is now different. Please use old logic.

Can you please explain? 
For me the current logic is quite the same:
1. StartInvoke
2. Wait normally for given timeout
3. If timed out -> Abort()
4.  Wait for Short timeout
5. If timed out -> Enter busy
6. Repeat from 4
7. If awaited -> Exit busy

nerzhulart:
>It doesn't take into account that the method may already be being aborted, like the old implementation did.

Fixed

PR #88, opened 2017-01-12
Evaluation fixes

nerzhulart:
Fixes
https://bugzilla.xamarin.com/show_bug.cgi?id=1637
and
https://youtrack.jetbrains.com/issue/RIDER-3767

### DIFF
--- a/Mono.Debugging.Soft/SoftEvaluationContext.cs
+++ b/Mono.Debugging.Soft/SoftEvaluationContext.cs
@@ -186,7 +186,7 @@ namespace Mono.Debugging.Soft
 			} catch (NotSupportedException) {
 				AssertTargetInvokeAllowed ();
 				var threadState = Thread.ThreadState;
-				if (threadState == ThreadState.WaitSleepJoin) {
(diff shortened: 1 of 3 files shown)


Mono.Debugging.Soft/SoftEvaluationContext.cs
```diff
@@ -184,7 +185,11 @@ Value RuntimeInvoke (MethodMirror method, object target, Value[] values, bool en
 				return method.Evaluate (target is TypeMirror ? null : (Value) target, values);
 			} catch (NotSupportedException) {
 				AssertTargetInvokeAllowed ();
-
+				var threadState = Thread.ThreadState;
+				if (threadState == ThreadState.WaitSleepJoin) {
```
DavidKarlas: This should be `if ((threadState & ThreadState.WaitSleepJoin) == ThreadState.WaitSleepJoin) {`

DavidKarlas:
Can you also add caching of ThreadState?

It should probably be invalidated at same time as frames in `InvalidateFrames` in ThreadMirror.cs.

PR #92, opened 2017-01-13
Handling exceptions in OnRun() before ForceExit

nerzhulart:
Handling exceptions in OnRun() before ForceExit, because otherwise the session is ended and the exception may be ignored.


DavidKarlas:
Why only in OnRun? There are many cases with:
```
ForceExit ();
if (!HandleException (ex))
     throw;
```

nerzhulart:
@DavidKarlas done

PR #99, opened 2017-02-09
Bug 52333 - Debugger applies rounding when casting float values to int

DavidKarlas:
Since System.Convert.To, can have different behavior then casting, I removed use of this and always use DynamicCast which does proper casting. In theory performance might suffer but this should be minimal and not noticable by user at all.

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
@@ -202,53 +202,60 @@ namespace Mono.Debugging.Soft
 			return null;
 		}
 
+		static Dictionary<string, TypeCastDelegate> typeCastDelegatesCache = new Dictionary<string, TypeCastDelegate> ();
 		static TypeCastDelegate GenerateTypeCastDelegate (string methodName, Type fromType, Type toType)
 		{
-			var argTypes = new [] { typeof (object) };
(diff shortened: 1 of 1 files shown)


slluis:
GenerateTypeCastDelegate is not doing any kind of caching of the generated method. That looks wrong.

DavidKarlas:
Added caching, it's based on method name which is good enough for all our use cases(casting between primary values)

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -202,53 +202,60 @@ public override object TryConvert (EvaluationContext ctx, object obj, object tar
...
-				PropertyInfo propInfo = fromType.GetProperty ("Value");
-				methodInfo = propInfo.GetGetMethod ();
-
-				il.Emit (OpCodes.Stloc_0);
-				il.Emit (OpCodes.Ldloca_S);
-				il.Emit (OpCodes.Call, methodInfo);
-
-				fromType = methodInfo.ReturnType;
-			}
+			lock(typeCastDelegatesCache) {
```
Therzok: NIt: missing space.

DavidKarlas: Formatting settings/Roslyn problem

PR #108, opened 2017-03-23
Fix handling of thread death for old protocol versions

jbevain:
Old debugger-agent servers can report a ThreadDeathEvent for an already collected ThreadMirror. In that case we should not prompt the user with the error.


jstedfast:
Seems reasonable to me, but adding David Karlas to the list of reviewers to confirm.

PR #115, opened 2017-04-11

joj:


joj:
This PR has an issue, closing for now.

PR #116, opened 2017-04-11

joj:


Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2640,58 +2640,67 @@ bool LoadPdbFile (string pdbFileName)
...
+
+				info.Hash = document.Hash;
+				info.FileID = 0;
+				info.FullFilePath = document.Url;
+
+				fileToSourceFileInfos.Add(name, new List<SourceFileDebugInfo> { info });
+			}
+			else
+			{
+				fileToSourceFileInfos[name].First().PdbMethods.Add(sequencePoints.ToArray());
```
Therzok: If all we have is one element in each value List, why do we still have a list?

The old code just appended a new SourceFileDebugInfo to the value, it didn't modify the value's contents.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2640,58 +2640,67 @@ bool LoadPdbFile (string pdbFileName)
...
+					AddSequencePointsToFileSource(fileToSourceFileInfos, doc.Document, doc.SequencePoints, Path.GetFileName(doc.Document.Url));
+				}
+			}
+		}
+
+		private static void AddSequencePointsToFileSource(Dictionary<string, List<SourceFileDebugInfo>> fileToSourceFileInfos, Document document, IEnumerable<SequencePoint> sequencePoints, string name)
+		{
+			if (!fileToSourceFileInfos.ContainsKey(name))
+			{
+				List<SequencePoint[]> list = new List<SequencePoint[]>(new[] { sequencePoints.ToArray() });
```
Therzok: Do we really need to do `ToArray()`?

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2640,58 +2640,67 @@ bool LoadPdbFile (string pdbFileName)
...
+			}
+
+			return true;
+		}
+
+		private void LoadPdbType(Cecil.TypeDefinition type, Dictionary<string, List<SourceFileDebugInfo>> fileToSourceFileInfos)
+		{
+			foreach (var method in type.Methods)
+			{
+				var documents = method.DebugInformation.SequencePoints.GroupBy(t => t.Document, t => t, (document, sequencePoints) => new { Document = document, SequencePoints = sequencePoints });
```
Therzok: Why are we grouping the documents? It doesn't seem like the code is dependent on this. We're using the same dictionary, so this is an operation that's not needed at all?

`SequencePoint` exposes document. I guess this is a left-over.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2640,58 +2640,67 @@ bool LoadPdbFile (string pdbFileName)
...
+			sourceFilesDebugInfo[pdbFileName] = fileToSourceFileInfos;
+
+
+			using (var module = Cecil.ModuleDefinition.ReadModule(Path.ChangeExtension(pdbFileName, ".dll")))
+			{
+				var symbolReaderProvider = new Cecil.Cil.DefaultSymbolReaderProvider(false);
+				module.ReadSymbols(symbolReaderProvider.GetSymbolReader(module, pdbFileName));
+
+				var methodMapping = new Dictionary<Document, List<SequencePoint[]>>();
+				foreach (var type in module.Types)
```
Therzok: Instead of manually doing this, please use `module.GetTypes()` which also returns `Nestedtypes`.

https://github.com/jbevain/cecil/blob/71ee20ba1e70c698f79901724facd2de31298329/Mono.Cecil/ModuleDefinition.cs

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2850,25 +2859,18 @@ bool CheckBetterMatch (TypeMirror type, string file, int line, int column, Locat
 							return true;
 						if (entry.Row == line && column >= entry.Column && entry.Column > found.ColumnNumber)
 							return true;
-						if (vm.Version.AtLeast (2, 19)) { //if version is less then 2.19, found Location will not contain info about columns
```
Therzok: Why was this code removed? Mdb handling should be unchanged?

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: I can change that. I didn't know that was the behavior of GetTypes but it's definitely what I need..

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: I'm grouping the documents to be able to iterate document instead of sequence points. It makes the code way more clear to read in the next few lines.

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: The end result is a list of array, so yes, I need the array. The next comment also on this.

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: Even though I agree on the general statement I didn't want to change existing code. The hileToSource is a dictionary of list of array, where the list must have only one item (there's a check in CheckBestMatch to see if the list has more than one item. As before, even though I think that's really not ideal I didn't want to change existing code too much. I can do it if it's preferred.

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: This was a request from @DavidKarlas. The previous line is identical to the one I removed, except for the vm check. The end result is that if the vm version check passes, the other will fail every time, so that whole thing was not needed. I can separate that check into a different commit for tracking purposes.

joj:
@Therzok I'll address the GetTypes thing tomorrow morning (it's late here right now). In the meantime, we can still discuss the others. In particular, I'm 100% ok with simplifying the dictionary of list of array, but that will require me to change mdb code too, which is something that you didn't want for the last one. I can split that into several commits for clarity and I will but that's out of scope for what I really need, which is to remove the System.Reflection.Metadata dll, as it is breaking our support for VS2013 and 2015.

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: Also, it lets me preserve the exact same contents the mdb generator creates, which avoids bugs down the line :)

joj:
I added a second commit with @Therzok's proposed changes. I will squash them before you guys merge if that's what he wanted.

joj:
debugger-libs is moving to cecil 0.9.6 (and has conflicts now) so closing this.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2640,58 +2640,62 @@ bool LoadPdbFile (string pdbFileName)
...
+			}
+		}
+
 		bool LoadMdbFile (string mdbFileName)
 		{
 			if (!File.Exists (mdbFileName))
 				return false;
 
-			var fileToSourceFileInfos = new Dictionary<string, List<SourceFileDebugInfo>> ();
+			var fileToSourceFileInfos = new Dictionary<string, List<SourceFileDebugInfo>> (StringComparer.InvariantCultureIgnoreCase);
```
mhutch: This should use the `PathComparer` property, surely?

joj: I think this is more an Id than a path. For starters, there's two copies of each (one with full path and one with just the file name), so I'm not sure how that will work :) Somebody else may know better.

PR #122, opened 2017-05-12

joj:


joj:
@Therzok this is what we discussed. Same as what's in the VS branch.

PR #124, opened 2017-05-19
Bug 53371 and domain breakpoints

DavidKarlas:


slluis:
LGTM, although it would be nice to have unit tests for this.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -1981,7 +1981,16 @@ void HandleAssemblyUnloadEvents (AssemblyUnloadEvent[] events)
 
 			if (assemblyLocation != null) {
 				foreach (var pair in source_to_type) {
-					pair.Value.RemoveAll (m => PathComparer.Equals (m.Assembly.Location, assemblyLocation));
+					pair.Value.RemoveAll (m => {
+						try {
+							assemblyLocation = m.Assembly.Location;
```
Therzok: wouldn't being able to set this yield true on all cases?

Therzok: What I mean is, this delegate will always return true, because we're setting assemblyLocation to m.Assembly.Location.

DavidKarlas: you are correct, it should be different variable, I will fix, tnx for spotting this

DavidKarlas:
WIll open new PR with tests...

PR #142, opened 2017-07-03
Limited Lambda Support on Immediate Pad

harukamm:
Support lambda expressions on expression evaluator.

#### Supported Features:
Lambdas satisfying the following conditions are supported.
* Lambdas with cast or method invocation
  - `((Action<string>)(x => System.Console.WriteLine (x))`
  - `(Func<int>)(() => 50 * 100)`
  - `lst.Find (x => x == “bar”)`
  - `arr.First (x => x == "foo")`
* Public type / public method / local variables access

#### Non-Supported Features:
1. Private type / private method access
2. Lambdas in some kinds of generic methods
    - Invocation of generic methods like `Method1<T> (T x, Func<T> f)` are supported because a generic type `T` can be resolved by the type of first parameter `x`.
    - However, this doesn’t work: `Method1<T> (Func<T> f)`
    - We have to provide type arguments like `Method1<int>(() => 50)`
3. Side Effect
    - `((Action<int>)(x => y = x)).Invoke(5)`, let y be one of local variables.
4. Async Lambdas

Note that my patches use Microsoft.CodeAnalysis ver1.3.2, while MonoDevelop references ver2.3.0-beta1. Ver-2.3.0 is not available under the target framework version 4.5. I'd like to ask reviewers if it's alright to update the target framework version.

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/DelayedLambdaType.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/DelayedLambdaType.cs
@@ -17,11 +17,13 @@ namespace Mono.Debugger.Soft
 		}
 
 		string expression;
+		Tuple<string, Value>[] locals;
 		string uniqueName;
 
-		public DelayedLambdaType (VirtualMachine vm, string expression) : base (vm, 0)
+		public DelayedLambdaType (VirtualMachine vm, Tuple<string, Value>[] locals, string expression) : base (vm, 0)
 		{
 			this.expression = expression;
+			this.locals = locals;
 			this.uniqueName = "Lambda" + Guid.NewGuid ().ToString ("N");
 		}
 
@@ -29,8 +31,24 @@ namespace Mono.Debugger.Soft
 			get { return expression; }
 		}
 
+		public Tuple<string, Value>[] Locals {
+			get {
+				if (locals == null)
+					return new Tuple<string, Value> [0];
+				return locals;
+			}
+		}
+
 		public string Name {
 			get { return uniqueName; }
+		}
+
+		public Value[] GetLocalValues ()
+		{
+			var vals = new Value [Locals.Length];
+			for (int i = 0; i < vals.Length; i++)
+				vals [i] = Locals [i].Item2;
+			return vals;
 		}
 
 		public bool IsAcceptableType (TypeMirror t)
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/DelayedLambdaType.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/DelayedLambdaType.cs
@@ -97,7 +97,7 @@ namespace Mono.Debugger.Soft
 			if (typ == null)
 				return null;
 
-			string fullName = typ.FullName;
+			string fullName = typ.FullName.Replace ('+', '.');
 			string nmespace = typ.Namespace == "" ? null : typ.Namespace;
 			string typeName = "";
 			string [] argTypeNames = null;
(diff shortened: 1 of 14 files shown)


Therzok:
Hey! This PR looks great! I'd love to see some positive/negative tests for the parsing and some tests for the evaluation itself.

jstedfast:
I'm not sure what, if any, implications there are for bumping the target framework version.

I would so love to see lambda support, though. 👍 

Hopefully we can bump it.

Therzok:
For MonoDevelop, it should have no implication. We bumped it already.

jstedfast:
Right, but MonoDevelop might not be the only consumer of debugger-libs

harukamm:
Thanks for your comments:) As for the target framework version, I'd like to defer decision to others because I'm new to this project.
I've added unit tests for this patches per suggestion from @Therzok.

Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
```diff
@@ -287,6 +287,7 @@ public virtual void OnBusyStateChanged (BusyStateEventArgs e)
 		public abstract bool IsEnum (EvaluationContext ctx, object val);
 		public abstract bool IsValueType (object type);
 		public abstract bool IsClass (EvaluationContext ctx, object type);
+		public abstract bool IsDelayedType (EvaluationContext ctx, object type);
```
Therzok: @DavidKarlas @jstedfast are we API stable? If so, we need to avoid adding new abstract members (use virtual) and changing abstract to non-abstract here.

Therzok:
I love where this is going! Thanks for the unit tests!

Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
harukamm: Done

Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
```diff
@@ -414,6 +415,8 @@ public virtual bool ForceLoadType (EvaluationContext ctx, object type)
 
 		public abstract object CreateValue (EvaluationContext ctx, object type, params object[] args);
 
+		public abstract object CreateDelayedLambdaValue (EvaluationContext ctx, string expression);
```
harukamm: Done

Mono.Debugger.Soft/Mono.Debugger.Soft/DelayedLambdaType.cs
```diff
@@ -0,0 +1,147 @@
...
+			//       argTypeNames: ["System.Int32", "System.Single"] }
+
+			string nameSpace = "";
+			string typeName = "";
+			string rest = fullName;
+			if (rest.StartsWith ("System.", StringComparison.Ordinal)) {
+				nameSpace = "System";
+				rest = rest.Substring (7);
+			} else {
+				throw new ArgumentException ("Currently, we support only lambda whose namespace is `System`");
```
harukamm: I set the limitation because this PR only supports System.Func typed lambda. My local version has already supported other lambdas, so this check was removed.

harukamm:
Thank you for the review!
@DavidKarlas and I talked offline, and we've decided to update the target framework version.

This patch is far behind my local change. It supports:

1. delegate type other than System.Func (e.g., Action/Predicate/public user-delegate)
2. access to outside the context

Should I close this PR and create new one, or push the changes to this PR?

DavidKarlas:
@harukamm can you fix merge conflict?

harukamm:
@DavidKarlas I've resolved the conflicts!

PR #145, opened 2017-07-10
Bug 57168 - No debugger support for named tuples

DavidKarlas:


Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -982,9 +996,28 @@ static T FindByName<T> (IEnumerable<T> items, Func<T,string> getName, string nam
...
+			return GetMembers (ctx, null, t, co, bindingFlags);
+		}
+
+		string [] GetTupleElementNames (IObjectSource source, EvaluationContext ctx)
+		{
+			switch (source) {
+			case FieldValueReference field:
+				return field.GetTupleElementNames ();
+			case PropertyValueReference prop:
+				return prop.GetTupleElementNames ();
```
slluis: If I'm following the code correctly, this will be called for every GetMember call on a property, and this ends calling TupleElementNamesPrivate in PortablePdbData, which opens and reads the pdb to get the tuple information. This seems to be a slow operation to do on every call.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -916,9 +916,15 @@ static bool IsAnonymousType (TypeMirror type)
 
 		protected override ValueReference GetMember (EvaluationContext ctx, object t, object co, string name)
 		{
+			return GetMember (ctx, null, t, co, name);
+		}
+
+		public override ValueReference GetMember (EvaluationContext ctx, IObjectSource objectSource, object t, object co, string name)
+		{
 			var type = t as TypeMirror;
+			var tupleNames = GetTupleElementNames (objectSource, ctx);
```
slluis: You should at least check that the type is a tuple before calling GetTupleElementNames.

DavidKarlas:
For Fields and Properties this information is not stored in .pdb but in Attributes, see changes in `FieldValueReference.cs` and `PropertyValueReference.cs`.
But for local variables this information is stored in .pdb...
Reason why this is code paths is not performance critical is that `GetTupleElementNames` is only called when object is expanded. So even if we have 30 local variables in method this method won't be called, unless user expands object to see fields and properties. And in that case it will be called only 1x(for expanded object). Anyway... Added optimisation to check for type name.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
DavidKarlas: Done.

Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
```diff
@@ -857,7 +857,7 @@ public ValueReference GetMember (EvaluationContext ctx, IObjectSource objectSour
 			return GetMember (ctx, objectSource, GetValueType (ctx, co), co, name);
 		}
 
-		public ValueReference GetMember (EvaluationContext ctx, IObjectSource objectSource, object t, object co, string name)
+		public virtual ValueReference GetMember (EvaluationContext ctx, IObjectSource objectSource, object t, object co, string name)
```
slluis: If you make this method virtual, the ParentSource assignment below won't be done if this member is overriden and base is not called. So please don't make this method virtual and create a new protected one instead that is called here.

Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -951,6 +957,14 @@ protected override ValueReference GetMember (EvaluationContext ctx, object t, ob
 			return null;
 		}
 
+		private static string MapTupleName (string name, string [] tupleNames)
+		{
+			if (tupleNames != null && name.Length > 4 && name.StartsWith ("Item", StringComparison.Ordinal) && int.TryParse (name.Substring (4), out var tupleIndex) && tupleNames.Length >= tupleIndex && tupleNames [tupleIndex - 1] != null)
```
Therzok: E_TOOWIDE. I'd format this differently to be a bit more readable.

Therzok: Also, redundant private.

UnitTests/MonoDevelop.Debugger.Tests.TestApp/TestEvaluation.cs
```diff
@@ -171,6 +171,10 @@ class TestEvaluation : TestEvaluationParent
 		string someString = "hi";
 		string[] numbers = { "one", "two", "three" };
 
+		public (int a, string, double B) ProperyA => (1, "2", 3.3);
```
Therzok: PropertyA not used in test. (also fix typo)

PR #148, opened 2017-08-02
Bug 57721 - Show Value on Popup non-functional in this case

DavidKarlas:
Issue was that those properties were introduced in macOs 10.13 and user ran 10.12
But bug was that we treated that same as if ImplicitEvaluation was disabled, hence showing "Show Value" button meant to force evaluation even with ImplicitEvaluation disabled... But when "Show Value" was clicked it still didn't evaluate because that was not allowed...

This is how it looks now:
image


DavidKarlas:
@joj How do you feel about Mono.Posix.dll addition used for localisation? Is this compatible with VS?

PR #150, opened 2017-08-03
Bug 33166 - Exception spew in the application output when writing in the immediate window

DavidKarlas:
Writing exceptions to "Application Output" caused by trying to get code completion makes no sense...

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
+++ b/Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
@@ -785,6 +785,7 @@ namespace Mono.Debugging.Evaluation
 					}
 
 					// FIXME: handle types and namespaces...
+				} catch (EvaluatorException) {
 				} catch (Exception ex) {
 					ctx.WriteDebuggerError (ex);
 				}


Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
```diff
@@ -785,8 +785,7 @@ public virtual CompletionData GetExpressionCompletionData (EvaluationContext ctx
 					}
 
 					// FIXME: handle types and namespaces...
-				} catch (Exception ex) {
-					ctx.WriteDebuggerError (ex);
+				} catch {
```
Therzok: Shouldn't we at least log them somewhere?

DavidKarlas: It will just be spew elsewhere, this exceptions are legit and OK exception, but not when trying to get CodeCompletion, in that case it doesn't make sense to show them...

slluis: If the evaluation method throws due to a parsing or evaluation error (EvaluatorException), then it is OK to ignore that exception. But if it throws due to some internal error, that exception should be logged.

If the NRefactory exceptions mentioned in the bug report are parsing or evaluation errors then they should not leak through the Evaluate method. They should be cached and and a corresponding EvaluatorException should be thrown instead.

DavidKarlas:
I updated to filter out only EvaluationExceptions but still log other exceptions. About .Evaluate shouldn't leak exceptions... This is how errors are propagated(https://github.com/mono/debugger-libs/blob/master/Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs?utf8=✓#L1368-L1377) via leaking exceptions

slluis:
I mean that it should not leak exceptions other than EvaluatorException and its subclasses.

PR #165, opened 2017-11-03
InvalidOperationException ("Already connecting") at second connection attempt 

VysotskiVadim:
I get **InvalidOperationException ("Already connecting")**, if first connection attempt fails.
Looks like it's because **0** is passed instead of **attemptNumber**.


mhutch:
This looks correct to me 👍 

/cc @DavidKarlas

PR #182, opened 2018-04-17
Add log points

iainx:
A log point is just a breakpoint that prints a message and continues


DavidKarlas:
Why do we need LogPoints when we already have TracePoints?

PR #185, opened 2018-05-30
Fix locals for local functions

mhutch:
The existing fix (#170) only handled the case where a local function captures locals from the parent scope on a ref struct. This fixes the case where the function does not capture and the case where it creates a class closure.

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
@@ -1070,11 +1070,23 @@ namespace Mono.Debugging.Soft
 				return GetHoistedLocalVariables (cx, vthis).Union (GetLocalVariables (cx));
 			}
 			if (IsLocalFunction (cx)) {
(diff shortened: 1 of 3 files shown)


Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -1070,12 +1070,25 @@ protected override IEnumerable<ValueReference> OnGetLocalVariables (EvaluationCo
 				return GetHoistedLocalVariables (cx, vthis).Union (GetLocalVariables (cx));
 			}
 			if (IsLocalFunction (cx)) {
-				//We are assuming here that last parameter is LocalFunction hoisted struct
-				var par = cx.Frame.Method.GetLocals ().Where (p => p.IsArg).Last ();
-				return GetHoistedLocalVariables (cx, new VariableValueReference (cx, par.Name, par)).Union (GetLocalVariables (cx));
+				ValueReference vthis = GetThisReference (cx);
```
mhutch: Oops, this line shouldn't be here

PR #193, opened 2018-08-18
Explicit Nuget.config. Update readme.md

VPeruS:
* Added explicit nuget.config
* Remove from readme that cecil must be cloned to same dir as debugger-libs

Without nuget.config msbuild on my system ArchLinux `Linux 4.17.6-1-ARCH` don't install external packages. Readme was changed because Cecil was added as package to .csproj files


nosami:
@VPeruS Thanks!

PR #196, opened 2018-10-01
SoftDebuggerAdaptor: support CoreFX Nullable implementation

akoeplinger:
When Mono switched to the Nullable implementation from CoreFX with https://github.com/mono/mono/commit/275d7bfc2ca17aa03a9daeb32493ee0ef9078037
some tests in VSMac broke because the SoftDebuggerAdaptor is checking for the old field name in the type.

Updated the adaptor to support both the CoreFX and the old Mono field name.

https://github.com/mono/mono/issues/10794


slluis:
@akoeplinger Does this affect 15.9?

marek-safar:
This was found during Mono 5.18 integration so it will only when someone uses that version or newer with 15.9

PR #198, opened 2018-11-16
Minor optimization - check for existence of .pdb before .mdb

nosami:
....as .mdb is deprectated

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -2758,11 +2758,11 @@ namespace Mono.Debugging.Soft
 
 			if (assemblyFileName == null)
 				return false;
-			string debugFile = assemblyFileName + ".mdb";
-			if (!LoadDebugFile (assemblyFileName, debugFile, LoadMdbFile)) {
-				debugFile = Path.ChangeExtension (assemblyFileName, ".pdb");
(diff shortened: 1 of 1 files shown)


joj:
We cannot do this in windows. pdb and mdb always exist side by side, so if we check for pdb first we'll never find mdb.

joj:
Also, we still have a lot of users using mdb in windows because of existing issues with forms that prevent them to migrate to ppdb.

joj:
@jstedfast see my comment before, but also for mdb it's doing (always was) `filename + ".mdb"` and for pdb it's doing `Path.ChangeExtension`, so the logic is ok.

jstedfast:
@joj right, that's what my comment was about - it's consistent with the way the code worked before, but I just thought the naming convention was AssemblyName.dll.pdb but I'm probably wrong :)

That said, if there's a reason we need to prefer mdb over pdb on Windows, then I guess this patch can't go in as-is.

Is there a reason mdb is preferred on Windows? Is it because the pdb and mdb files are inconsistent? e.g. maybe the pdb files are from the Windows-side build and the mdb's are from the Mac-agent portion of the build and are thus inconsistent?

joj:
It's not that we prefer mdb, it's just that this commit in windows completely prevents the usage of mdb, because the pdb will _always_ be there. We do have a lot of users still using mdb.

jstedfast:
Oh, okay.

PR #211, opened 2019-01-10
[Debugger] Record time it took between steps.

thaystg:
I implemented a new message to avoid breaking the protocol. I used the MonoStopwatch to count the time between the steps, it was already used in the debugger-agent.
Changing the method OnGetElapsedTime from abstract to virtual.
Changing the return to -1 to identify when the elapsed time is not available.
Fixes #8460


slluis:
Can you please also document that the returned value is in milliseconds?

thaystg:
> Can you please also document that the returned value is in milliseconds?

Done!

slluis:
Thanks!

PR #214, opened 2019-01-26
Move to PackageReference and 4.7.2.

KirillOsenkov:


Mono.Debugger.Soft/Mono.Debugger.Soft.csproj
```diff
@@ -118,22 +118,12 @@
   <ItemGroup>
     <None Include="Makefile.am" />
     <None Include="mono-git-revision" />
-    <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Mono.Cecil" Version="0.10.1" PrivateAssets="all" />
```
sandyarmstrong: Anyway way to get these using `$(NuGetVersionCecil)` when built in MonoDevelop? Is that worthwhile or no? I'm not sure how many ways these are built/distributed.

KirillOsenkov: I've added PrivateAssets="all" so it doesn't get inherited by referencing projects.

Therzok: @KirillOsenkov can we use the same patetrn as in RefactoringEssentials?

https://github.com/mono/RefactoringEssentials/blob/monodevelop2/RefactoringEssentials/RefactoringEssentials.csproj#L24-L26

KirillOsenkov: good idea, will do.

PR #216, opened 2019-04-16
[Mono.Debugging.Soft] support building on Windows

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/commit/bf58a880e5da244462fbae401f06153569836c96

In xamarin-android, we have started using this library for some of our
integration tests.

This causes our Windows build to fail with:

    SoftDebuggerSession.cs(2606,30): error CS0234: The type or namespace name 'Unix' does not exist in the namespace 'Mono' (are you missing an assembly reference?)
    [C:\src\xamarin-android\external\debugger-libs\Mono.Debugging.Soft\Mono.Debugging.Soft.csproj]

For now, we just need these libraries to *build* on Windows.

So if we conditionally use a NuGet package for Mono.Posix:

    <Reference        Condition=" '$(OS)' != 'Windows_NT' " Include="Mono.Posix" />
    <PackageReference Condition=" '$(OS)' == 'Windows_NT' " Include="Mono.Posix-4.5" Version="4.5.0" />

I'm able to build on Windows, and all is well.

### DIFF
--- a/Mono.Debugging.Soft/Mono.Debugging.Soft.csproj
+++ b/Mono.Debugging.Soft/Mono.Debugging.Soft.csproj
@@ -75,7 +75,8 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Core" />
-    <Reference Include="Mono.Posix" />
+    <Reference        Condition=" '$(OS)' != 'Windows_NT' " Include="Mono.Posix" />
+    <PackageReference Condition=" '$(OS)' == 'Windows_NT' " Include="Mono.Posix-4.5" Version="4.5.0" />
   </ItemGroup>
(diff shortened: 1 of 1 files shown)


joj:
As long as this doesn't pull posix into the vsix for the mono debugger, I'm ok with it. We used to provision posix externally to be able to build, but that was before nuget was used for everything :)

jonathanpeppers:
@joj this change does create:
```
Mono.Debugging.Soft\bin\Debug\Mono.Posix.dll
```
Where can I look at the VSIX? surely it doesn't wildcard with `*.dll`?

joj:
Don't worry, I'll take care of this when I bump. We are already excluding some files, it shouldn't be hard.

PR #218, opened 2019-05-03
Modified Breakpoints lock to be unique to breakpointStore

joj:
breakpointStore is using a shared lock (slock) that makes it lock when anything is happening. The problem is that if an action is happening at that time Breakpoints will lock until that is completed. This is being reported as one of the major UI locks in VS right now. I'm modifying the lock to be unique to breakpointStore. The field is only being accessed via the property (except on OnStarted, where I modified it to be accessed via the property) so breakpointStore will always be locked when accessed or modified.


joj:
To be replaced with something else.

PR #220, opened 2019-05-08
Added DotNetCore test projects

jstedfast:


jstedfast:
This is still a work-in-progress as I'm not even sure these projects will build.

jstedfast:
This is now working and is needed for https://github.com/mono/monodevelop/pull/7531

Mono.Debugging/Mono.Debugging.Evaluation/ObjectValueAdaptor.cs
```diff
@@ -255,7 +255,7 @@ static int FindTypeEnd (string typeName, int startIndex, int endIndex)
 			return i;
 		}
 		
-		public virtual string GetShortTypeName (string typeName)
+		public static string GetShortTypeName (string typeName)
```
Therzok: API break? Important?

jstedfast: Yea, it's an API break but I don't think we should care. I mean, I suppose that I could put that static method somewhere else and reintroduce the virtual method which calls the static utility function...

Therzok: If nothing breaks (cc @jbevain), sounds good to me.

Also @kzu since you guys had something on top of this, I think

jstedfast: No need for @kzu or @jbevain to comment anymore... I've added back the virtual method and added a new static method that the virtual method calls by default. No more API breakage.

PR #227, opened 2019-06-06
[Mono.Debugging] Improved source checksum logic for UNIX and DOS line…

jstedfast:
… endings

Since it is possible that the assemblies being debugged could have
been compiled on a different system (potentially with different line
endings), compute hashes for the source code in 3 formats:

1. The source code file with line endings as-is
2. The source code file with DOS line endings
3. The source code file with UNIX line endings

Note: sometimes source code may have mixed line endings, so it is
important to calculate the source code as-is and not just
canonicalized UNIX and DOS line endings.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/642329
Fixes issue #5260

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/SourceLocation.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/SourceLocation.cs
@@ -1,4 +1,7 @@
 using System;
+using System.IO;
+using System.Buffers;
+using System.Security.Cryptography;
 
 namespace Mono.Debugging.Client
 {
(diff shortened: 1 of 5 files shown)


Mono.Debugging/Mono.Debugging.Client/SourceLocation.cs
```diff
@@ -35,5 +38,124 @@ public override string ToString ()
...
+						pc = c;
+					}
+
+					hash.TransformBlock (buffer, 0, count, outputBuffer: null, outputOffset: 0);
+					dos.TransformBlock (dosBuffer, 0, dosIndex, outputBuffer: null, outputOffset: 0);
+					unix.TransformBlock (unixBuffer, 0, unixIndex, outputBuffer: null, outputOffset: 0);
+				}
+
+				hash.TransformFinalBlock (buffer, 0, 0);
+				dos.TransformFinalBlock (buffer, 0, 0);
```
Therzok: should this be on dosBuffer?

PR #231, opened 2019-06-21
Added EvaluationContext to AsyncOperation

jstedfast:
SoftDebugger's MethodCall async op already had it (Cor did not),
but we need this to report telemetry on what evaluation options
were enabled for the particular async op that is busy.

Partial "fix" to hopefully allow us to get more info about why/when
this is happening for our users.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802365/

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
+++ b/Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
@@ -34,7 +34,7 @@ namespace Mono.Debugging.Evaluation
 {
 	public class AsyncOperationManager: IDisposable
 	{
-		List<AsyncOperation> operationsToCancel = new List<AsyncOperation> ();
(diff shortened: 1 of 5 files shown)


Mono.Debugging/Mono.Debugging.Evaluation/AsyncOperationManager.cs
```diff
@@ -49,16 +49,19 @@ public void Invoke (AsyncOperation methodCall, int timeout)
...
 					lock (operationsToCancel) {
 						operationsToCancel.Remove (methodCall);
 						ST.Monitor.PulseAll (operationsToCancel);
 					}
+
 					if (wasAborted)
 						throw new EvaluatorAbortedException ();
-					else
-						throw new TimeOutException ();
+
```
sgmunn: did you mean to remove the else?

PR #235, opened 2019-07-19
Sourcelink

nosami:

PR #236, opened 2019-07-22
Fix tests

nosami:
This should have been added [here](https://github.com/mono/debugger-libs/pull/235) really but I already merged it :(


jstedfast:
I actually already commented out those blocks last week, but it would probably be better to remove them (like you did).

Seems there's a merge conflict because of that.

nosami:
@jstedfast ah, didn't realise.... I'll rebase

PR #238, opened 2019-07-26
Sourcelink

nosami:


Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -153,7 +162,7 @@ SourceLinkMap[] GetSourceLinkMaps ()
 
 		internal SoftScope [] GetHoistedScopesPrivate (MethodMirror method)
 		{
-			using (var fs = new FileStream (pdbFileName, FileMode.Open, FileAccess.Read, FileShare.Read))
+			using (var fs = stream)
 			using (var metadataReader = MetadataReaderProvider.FromPortablePdbStream (fs)) {
```
mrward: Might be a dumb question here, but what happens if say GetSourceLinkMaps is called and then GetHoistedScopesPrivate is called? The stream is disposed by GetSourceLinkMaps - will that result in an exception? Just wondering since the original code creates a new file stream each time.

sgmunn: @mrward from what I quickly saw from the code that uses it in Mono.Debugger is that the code pattern is

```
    ctx.Session.GetPdbData (variable.Method)?.XXXXX
```

where `GetPdbData` returns a new instance of `PortablePdbData` each time, so it does not look like we hit this but it does seem like a valid concern. Essentially we can use this object just the once if we hit a method that uses stream which doesn't sound entirely right @jstedfast @nosami

nosami: Yeah, this code is not very good. I've found a better way to do the same thing, so I'm going to rewrite this part.

nosami: The better way to do this isn't supported on latest Android, so going to need to refactor.

nosami: The fact that it returns a new instance of `PortablePdbData` each time needs changing to return a cached version, so the stream issue would become a bigger concern

nosami: Fixed

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -130,8 +130,8 @@ public SourceLink GetSourceLink (string originalFileName)
 
 		SourceLinkMap[] GetSourceLinkMaps ()
 		{
-			using (var fs = stream)
-			using (var provider = MetadataReaderProvider.FromPortablePdbStream (fs)) {
+			using (stream)
+			using (var provider = MetadataReaderProvider.FromPortablePdbStream (stream)) {
```
Therzok: By default, `FromPortablePdbStream` disposes of the underlying stream. `using (stream)` seems unneeded.

Do we need to pre-fetch metadata to avoid later IO?

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -77,73 +80,30 @@ internal class SoftScope
...
-			public string UriWildcard { get; }
-
-			public SourceLinkMap (string relativePathWildcard, string uriWildcard)
-			{
-				UriWildcard = uriWildcard;
-				RelativePathWildcard = relativePathWildcard;
-			}
-		}
+			using (var stream = GetStream()) {
+				using (var provider = MetadataReaderProvider.FromPortablePdbStream (stream)) {
```
Therzok: By default, it closes the underlying stream, unless `LeaveOpen` is passed. Should we `pre-fetch metadata` though?

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+			var pdbBlob = asm.GetPdbBlob ();
+			return pdbBlob != null ? new PortablePdbData (pdbBlob) : null;
 		}
 
 		internal PortablePdbData GetPdbData (MethodMirror method)
 		{
 			return GetPdbData (method.DeclaringType.Assembly);
 		}
 
+		Dictionary<AssemblyMirror, SourceLinkMap []> SourceLinkCache = new Dictionary<AssemblyMirror, SourceLinkMap []> ();
```
Therzok: Should we actually use a ConditionalWeakTable here? We want the SourceLinkMap to last as much as the AssemblyMirror, so if the AssemblyMirror does end up collected, we want the SourceLinkMap to be collected too?

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -62,12 +59,18 @@ public static bool IsPortablePdb (string pdbFileName)
...
 		}
 
-		private string pdbFileName;
+
+		readonly byte [] pdbBytes;
+		private readonly string pdbFileName;
 
 		public PortablePdbData (string pdbFileName)
 		{
 			this.pdbFileName = pdbFileName;
```
Therzok: This will handle some exceptions on invalid paths, helping us handle invalid data better.

```suggestion
			this.pdbFileName = Path.GetFullPath (pdbFileName);
```

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -62,12 +59,18 @@ public static bool IsPortablePdb (string pdbFileName)
...
-			sourceLinkMaps = new Lazy<SourceLinkMap[]> (GetSourceLinkMaps);
+		}
+
+		public PortablePdbData (byte [] pdbBytes)
+		{
+			this.pdbBytes = pdbBytes;
```
Therzok:
```suggestion
			this.pdbBytes = pdbBytes ?? throw new ArgumentNullException (pdbBytes);
```

Mono.Debugging.Soft/PortablePdbData.cs
Therzok: These are now needed, as we have 2 constructors

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -77,73 +80,30 @@ internal class SoftScope
...
-					// org/project-name/git-sha (usually)
-					var pathAndQuery = new Uri (httpBasePath).GetComponents (UriComponents.Path, UriFormat.SafeUnescaped).Substring (1);
-					// org/projectname/git-sha/path/to/file.cs
-					var relativePath = Path.Combine (pathAndQuery, localPath);
-					// Replace something like "f:/build/*" with "https://raw.githubusercontent.com/org/projectname/git-sha/*"
-					var httpPath = Regex.Replace (originalFileName, pattern, httpBasePath);
-					return new SourceLink (httpPath, relativePath);
-				}
+					var jsonBlob =
+						(from cdiHandle in pdbReader.GetCustomDebugInformation (EntityHandle.ModuleDefinition)
```
Therzok: Can we avoid LINQ? I think LINQ extensions make it easier to reason about captures.

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -77,73 +80,30 @@ internal class SoftScope
...
+						 let cdi = pdbReader.GetCustomDebugInformation (cdiHandle)
+						 where pdbReader.GetGuid (cdi.Kind) == SourceLinkGuid
+						 select pdbReader.GetBlobBytes (cdi.Value)).FirstOrDefault ();
```
Therzok: `pdbReader.GetGlobReader (cdi.Value).ReadSerializedString()` to avoid UTF8 APIs and multiple serialization

Mono.Debugging.Soft/PortablePdbData.cs
```diff
@@ -153,8 +113,8 @@ SourceLinkMap[] GetSourceLinkMaps ()
 
 		internal SoftScope [] GetHoistedScopesPrivate (MethodMirror method)
 		{
-			using (var fs = new FileStream (pdbFileName, FileMode.Open, FileAccess.Read, FileShare.Read))
-			using (var metadataReader = MetadataReaderProvider.FromPortablePdbStream (fs)) {
+			using (var stream = GetStream())
+			using (var metadataReader = MetadataReaderProvider.FromPortablePdbStream (stream)) {
```
Therzok: Not necessary, but would be good to do. Instead of reading the bytes, we could get the blobreader and read integer by integer.

Read a compressed integer for the length, and then read 2 integer values for the values we're looking for.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+		SourceLinkMap [] GetSourceLinkMaps (MethodMirror method)
+		{
+			var asm = method.DeclaringType.Assembly;
+			SourceLinkMap [] maps = Array.Empty<SourceLinkMap>();
```
Therzok:
```suggestion
```

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+
+			if (SourceLinkCache.TryGetValue (asm, out maps)) {
```
Therzok:
```suggestion
			if (SourceLinkCache.TryGetValue (asm, out SourceLinkMap[] maps)) {
```

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+					var jsonSourceLink = JsonConvert.DeserializeObject<JsonSourceLink> (jsonString);
+
+					if (jsonSourceLink != null && jsonSourceLink.Maps != null && jsonSourceLink.Maps.Any ())
+						maps = jsonSourceLink.Maps.Select (kv => new SourceLinkMap (kv.Key, kv.Value)).ToArray ();
+				} catch (JsonException ex) {
+					DebuggerLoggingService.LogError ("Error reading source link", ex);
+				}
+			}
+			
+			SourceLinkCache.Add (asm, maps);
```
Therzok: What's the threading story here? If 2 threads start computing the cache at once when it's empty, they can trigger an exception here:
```
t1 - no cache
t2 - no cache
t1 - add map
t2 - add map -> exception
```

Therzok: We either need a lock on the dictionary or we need to make this a:

```suggestion
			SourceLinkCache [asm] = maps;
```

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+		SourceLinkMap [] GetSourceLinkMaps (MethodMirror method)
+		{
+			var asm = method.DeclaringType.Assembly;
+			SourceLinkMap [] maps = Array.Empty<SourceLinkMap>();
+
+			if (SourceLinkCache.TryGetValue (asm, out maps)) {
+				return maps;
+			}
+
+			string jsonString = null;
```
Therzok: The code assumed the maps array would remain non-null after TryGetValue is done, but it would override it on failure. In case we would try to add a `null` to the map below.

```suggestion
			maps = Array.Empty<SourceLinkMap> ();
			string jsonString = null;
```

Therzok: Fix for this: https://github.com/mono/debugger-libs/pull/238/files#r310340014

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+			return maps;
+		}
+
+		internal SourceLink GetSourceLink (MethodMirror method, string originalFileName)
+		{
+			if (originalFileName == null)
+				return null;
+
+			var maps = GetSourceLinkMaps (method);
+			if (maps == null)
```
Therzok: We also return an empty array here. Maybe work checking that too

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -737,19 +739,84 @@ protected override ProcessInfo[] OnGetProcesses ()
...
+				return null;
+
+			originalFileName = originalFileName.Replace ('\\', '/');
+			foreach (var map in maps) {
+				var pattern = map.RelativePathWildcard.Replace ("*", "").Replace ('\\', '/');
```
Therzok: What about non-unix?

Mono.Debugging.Soft/SourceLinkMap.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Mono.Debugging.Soft
+{
+	public class SourceLinkMap
```
Therzok:
```suggestion
	internal class SourceLinkMap
```

Therzok:
Done with review pass 1.

Mono.Debugging.Soft/SoftDebuggerSession.cs
nosami: I don't think it matters, AssemblyMirror and DebuggerSession have the same lifespan.

Mono.Debugging.Soft/SoftDebuggerSession.cs
nosami: Combined with the previous suggestion, maps can now be null? I was ensuring that it could never be null.

Mono.Debugging.Soft/SoftDebuggerSession.cs
nosami: I was already ensuring that maps could not be null. The Indexer suggestion is a good one though

nosami: I guess I'll change it anyway, maybe it's clearer your way

Mono.Debugging.Soft/PortablePdbData.cs
nosami: > Should we pre-fetch metadata though?

This might fetch more data than we need. I think this option is for users who want to have full control over the stream after reading.

Mono.Debugging.Soft/PortablePdbData.cs
nosami: > Do we need to pre-fetch metadata to avoid later IO?

This might fetch more data than we need. I think this option is for users who want to have full control over the stream after reading.

Mono.Debugging.Soft/PortablePdbData.cs
nosami: This was existing code

Mono.Debugging.Soft/SoftDebuggerSession.cs
nosami: I'm not understanding this suggestion

Mono.Debugging.Soft/PortablePdbData.cs
Therzok: Gotcha. Maybe in a future iteration. If you want, I can file a VSTS ticket

Mono.Debugging.Soft/SoftDebuggerSession.cs
Therzok: Tl;dr:

Variable is set to Array.Empty.
Dictionary.TryGetValue will reset it to null.
If no mapping is found, it'll remain nu, thus NRE

Therzok: Due to now doing `maps ?? Array.Empty` the bug is no longer there.

I thought we weren't null checking return values, so it would've ended up in NRE.

Mono.Debugging.Soft/PortablePdbData.cs
nosami: Yeah, please

Mono.Debugging.Soft/PortablePdbData.cs
nosami: Your suggestion looked good to me. Unfortunately, it doesn't quite work - the json is truncated
```
"documents":{"D:\\a\\1\\s\\*":"https://raw.githubusercontent.com/xamarin/Xamarin.Forms/d7eda5af499126718262dbbd1ddd2b54b199
```

Therzok: That's weird. It should properly deserialize to a string.

Mono.Debugging.Soft/PortablePdbData.cs
Therzok: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/960770

Mono.Debugging.Soft/SoftDebuggerSession.cs
Therzok: Any out var is going to be assigned by the caller. So regardless of whatever is before it, it is going to be assigned (most likely to the default) by the callee.

nosami: Yep, I see now. Fixed

nosami:
Wow... Think I should stop using that automatic suggestion merge feature https://github.com/mono/debugger-libs/pull/238/commits/9cc5561cf0fd74fb11a0ee423f38c8c3721b0123

Therzok:
I'll stop using suggestions when it breaks the code.

Sadly, no way to know if the file is going to get changed line endings.

PR #239, opened 2019-07-31
Nullcheck GetPdbData for Sourcelink

sgmunn:


nosami:
@sgmunn looks a lot like this commit :) https://github.com/mono/debugger-libs/pull/238/commits/a2a5c6c8007936ef0c62012b58cb4a18a78ef305

sgmunn:
sure does, but what branch is that in? does not seem to be in master for some reason.

PR #240, opened 2019-07-31
Make BreakpointStore thread safe

sgmunn:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/949284

This adds locks around the manipulation of the breakpoints in the breakpoint store and ensures that once a caller gets a copy of the breakpoints that the collection is not modified afterwards by other threads. Adds a simple test to validate - which does fail with the prior implementation.

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
@@ -351,12 +351,12 @@ namespace Mono.Debugging.Client
 
 		public IEnumerator GetEnumerator ()
 		{
-			return InternalGetBreakpoints ().ToList().GetEnumerator ();
+			return ((IEnumerable)InternalGetBreakpoints()).GetEnumerator ();
 		}
 
 		IEnumerator<BreakEvent> IEnumerable<BreakEvent>.GetEnumerator ()
 		{
-			return InternalGetBreakpoints ().ToList ().GetEnumerator ();
+			return ((IEnumerable<BreakEvent>)InternalGetBreakpoints ()).GetEnumerator ();
 		}
 
 		public void Clear ()
--- a/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
@@ -476,7 +476,15 @@ namespace Mono.Debugging.Client
 			try {
 				buffer = Marshal.AllocHGlobal (PATHMAX);
 				var result = realpath (path, buffer);
-				return result == IntPtr.Zero ? "" : Marshal.PtrToStringAuto (buffer);
+				var realPath = result == IntPtr.Zero ? "" : Marshal.PtrToStringAuto (buffer);
+
+				if (string.IsNullOrEmpty (realPath) && !File.Exists (path)) {
+					// if the file does not exist then `realpath` will return empty string
+					// default to what we would do if calling this on windows
+					realPath = Path.GetFullPath (path);
+				}
+
+				return realPath;
 			} finally {
 				if (buffer != IntPtr.Zero)
 					Marshal.FreeHGlobal (buffer);
--- a/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
@@ -33,6 +33,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Runtime.InteropServices;
+using System.Collections.Immutable;
 
 namespace Mono.Debugging.Client
 {
@@ -41,6 +42,9 @@ namespace Mono.Debugging.Client
 		static readonly StringComparer PathComparer;
 		static readonly bool IsWindows;
 		static readonly bool IsMac;
+
+		readonly object breakpointLock = new object ();
+		ImmutableArray<BreakEvent> breakpoints = ImmutableArray.Create<BreakEvent>();
 
 		static BreakpointStore ()
 		{
@@ -73,21 +77,17 @@ namespace Mono.Debugging.Client
 			}
 			return false;
 		}
-
-		readonly List<BreakEvent> breakpoints = new List<BreakEvent> ();
 		
 		public int Count {
 			get {
-				return breakpoints.Count;
+				return InternalGetBreakpoints().Length;
 			}
 		}
 
 		public bool IsReadOnly {
 			get {
 				var args = new ReadOnlyCheckEventArgs ();
-				var checkingReadOnly = CheckingReadOnly;
-				if (checkingReadOnly != null)
-					checkingReadOnly (this, args);
+				CheckingReadOnly?.Invoke (this, args);
 				return args.IsReadOnly;
 			}
 		}
@@ -135,8 +135,11 @@ namespace Mono.Debugging.Client
 			if (IsReadOnly)
 				return false;
 
-			breakpoints.Add (bp);
-			bp.Store = this;
+			lock (breakpointLock) {
+				SetBreakpoints (breakpoints.Add (bp));
+				bp.Store = this;
+			}
+
 			OnBreakEventAdded (bp);
 
 			return true;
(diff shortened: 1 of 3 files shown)


Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
```diff
@@ -476,7 +476,15 @@ static string ResolveFullPath (string path)
 			try {
 				buffer = Marshal.AllocHGlobal (PATHMAX);
 				var result = realpath (path, buffer);
-				return result == IntPtr.Zero ? "" : Marshal.PtrToStringAuto (buffer);
+				var realPath = result == IntPtr.Zero ? "" : Marshal.PtrToStringAuto (buffer);
+
```
jstedfast: I wonder if we could clean this up a bit by doing:

```csharp
if (result == IntPtr.Zero && !File.Exists (path))
```

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
```diff
@@ -73,21 +77,17 @@ static bool IsRunningOnMac ()
...
 			return false;
 		}
-
-		readonly List<BreakEvent> breakpoints = new List<BreakEvent> ();
 		
 		public int Count {
 			get {
-				return breakpoints.Count;
+				return InternalGetBreakpoints().Length;
 			}
```
mrward: Not sure we need to lock to check the length of an ImmutableArray.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
```diff
@@ -572,14 +615,31 @@ public void FileRenamed (string oldPath, string newPath)
...
+			var oldEvents = breakpoints;
+			breakpoints = newBreakpoints;
+			return oldEvents;
+		}
+
+		ImmutableArray<BreakEvent> InternalGetBreakpoints ()
+		{
+			lock (breakpointLock) {
+				return breakpoints;
+			}
```
mrward: Locking access to an ImmutableArray does not look right to me. It is a thread safe type. Locking a non-thread safe collection such as a List would make sense.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: No, not necessary.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: Agreed. That lock is unnecessary.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
```diff
@@ -326,12 +351,12 @@ public ReadOnlyCollection<Breakpoint> GetBreakpointsAtFileLine (string filename,
 
 		public IEnumerator GetEnumerator ()
 		{
-			return breakpoints.GetEnumerator ();
+			return InternalGetBreakpoints ().ToList().GetEnumerator ();
```
slluis: Why ToList? It is not necessary. Same below.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: A bit more explanation. Calling InternalGetBreakpoints will return a breakpoint list. This list can be the one that exists before calling InternalGetBreakpoints, or a new one if new breakpoints are added or removed during the execution of InternalGetBreakpoints. The lock doesn't really have any effect on which list will be returned, since the change could be done at any time, before the lock is taken or after. Also, it is safe to access the breakpoints properties without a lock, since the reference assignment is atomic.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: `breakpoints` is an ImmutableArray, so by nature it can't change. It is totally safe to return an enumerator for it, there is no need to create a temporary copy with ToList()

sgmunn: Because of 

```
/Users/gregm/Projects/vsm/a/monodevelop/main/external/debugger-libs/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs(11,11): Error CS0029: Cannot implicitly convert type 'System.Collections.Immutable.ImmutableArray<Mono.Debugging.Client.BreakEvent>.Enumerator' to 'System.Collections.IEnumerator' (CS0029) (Mono.Debugging)
```

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
sgmunn: this "Also, it is safe to access the breakpoints properties without a lock, since the reference assignment is atomic." was what I was curious about and thought that might have been but when I discussed with Marius to check my understanding he did not think it was.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
sgmunn: Only for the reason below about atomicity.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: Reference assignments are certainly atomic. The lock is not necessary.

sgmunn: `breakpoints` itself is thread safe, what I wanted to validate by asking Marius was if the read of the breakpoints field would be guaranteed to be atomic because the field value could change from any other thread whilst reading. Perhaps we were not on the same page at the time, but that is the reason for the locks on reading.

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: Even though GetEnumerator() is not exposed as a public method, ImmutableArray does implement IEnumerator, so  you can do: `((IEnumerable)breakpoints).GetEnumerator()`

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
slluis: Ok, the problem is that ImmutableArray is not a reference type, it is a struct, and value type assignments are not always thread safe. Let me find out. The lock might be necessary after all!

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
sgmunn: I don't think it does implement IEnumerator. It does implement IEnumerable but I cannot see it implementing IEnumerator.

slluis: I mean that it implements IEnumerable, sorry, so you can get an enumerator from it.

sgmunn: 🤦‍♂

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
sgmunn: I think that depends on whether realpath can return Intptr.Zero and the file does exist. In that case we'd try to marshal a string from a zero pointer. the current code handles that case presently (returning an empty string).

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
Therzok: ImmutableArray assignment is most likely not atomic, hence why [InterlockedExchange](https://source.dot.net/#System.Collections.Immutable/System/Collections/Immutable/ImmutableInterlocked.cs,110) exists

Therzok: Which makes use of a nifty trick, and that is atomically swapping the underlying arrays of the structs.

slluis: Given that ImmutableArray<T> has one single field that is an array reference, I'm pretty sure assignment is atomic. InterlockedExchange does more than an assignment, it reads the value and assigns the new value all in an atomic operation.

PR #246, opened 2019-08-30
Ignore invocations count test on .NET Core debugger

rodrmoya:


UnitTests/Mono.Debugging.Tests/Shared/AdvancedEvaluationTests.cs
```diff
@@ -227,6 +227,7 @@ public void YieldMethodTest ()
 		[Test]
 		public void InvocationsCountDuringExpandingTest ()
 		{
+			IgnoreCorDebugger ("Behavior has changed in CoreClr debugger 1-21-0");
```
nosami: IgnoreVsDebugger

PR #249, opened 2019-09-13
Modernized SoftDebuggerAdaptor.cs

jstedfast:


Mono.Debugging.Soft/SoftDebuggerAdaptor.cs
```diff
@@ -635,25 +625,25 @@ private Compilation CreateLibraryCompilation (SoftEvaluationContext ctx, string
 
 		private byte [] CompileLambdaExpression (SoftEvaluationContext ctx, DelayedLambdaType val, string typeName, out string error)
 		{
-			string className = val.Name;
-			string assemblyNamePrefix = "lambdaAssem";
-			string assemblyName = assemblyNamePrefix + className;
-			string lambdaExpression = val.Expression;
+			var className = val.Name;
```
Therzok: Not too keen on these changes. Due to editorconfig not properly setup, we don't have the 'var when type is not clear' setting on.

PR #250, opened 2019-09-23
New way to filter exceptions to support VSWin features

thaystg:
Creating a new version of MOD_KIND_EXCEPTION_ONLY to support Not Filtered Exception feature, discussed with Joaquin Jares.

Because we don't have this functionality of get every other exception that is not specified, VsWin was getting every exception and filtering later. But this causes a slowness during the debugger because we stop all thread on each exception that we get, even if they are caught and we don't want to stop.

This fix a bug of slowness on Android Debugger.

Done together with this PR: https://github.com/mono/mono/pull/16825

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -305,6 +305,12 @@ namespace Mono.Debugger.Soft
 		public bool Subclasses {
 			get; set;
 		}
+		public bool NotFilteredFeature {
+			get; set;
+		}
+		public bool EverythingElse {
+			get; set;
+		}
 	}
 
 	class AssemblyModifier : Modifier {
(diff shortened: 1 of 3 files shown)


jstedfast:
Normally what we do is wait for things to be accepted into mono and then sync the changes over to debugger-libs

thaystg:
Thanks to let me know, I will wait.

jstedfast:
No problem :-)

thaystg:
@jstedfast , can we merge now?

Therzok:
Not all changes from the mono PR are in this PR

thaystg:
> Not all changes from the mono PR are in this PR

I've checked it again and I don't see what I've missed, could you tell me, please?

jstedfast:
Didn't you make changes to your original commit in mono? Did those updates get ported over to here?

thaystg:
Yes

Therzok:
Yeah, I re-checked the PRs and then deleted the comment as the changes were backported already, sorry!

PR #251, opened 2019-09-24
Revamped expression resolver to have context

jstedfast:
This is needed to resolve ambiguous expressions such as
those that could be a type -or- a variable name.

Partial fix for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/939728/


jstedfast:
Merging this will cause monodevelop to fail to compile until https://github.com/mono/monodevelop/pull/8800 is complete and gets merged...

PR #253, opened 2019-10-11
[d16-4][Mono.Debugging.Evaluation] Set the expression as the ObjectValue.Name

jstedfast:
This is what the old Gtk-based ObjectValueTreeView was doing
at a higher level, but that was always a hack.

This is a better way of doing it.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/993192/


jstedfast:
This is a backport of https://github.com/mono/debugger-libs/pull/252

PR #255, opened 2019-10-17
Exposed ExceptionEventRequest

joj:
In Visual Studio in Windows there is no concept of a Catchpoint. I'm exposing the Exception Event Requests so we can use them directly.

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -594,6 +594,8 @@ namespace Mono.Debugging.Soft
 
 			sourceFilesDebugInfo.Clear ();
 
+			exceptionRequests.Clear();
+
 			if (!HasExited) {
 				if (vm != null) {
 					ThreadPool.QueueUserWorkItem (delegate {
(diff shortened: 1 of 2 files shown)


Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -1212,20 +1215,88 @@ void InsertCatchpoint (Catchpoint cp, BreakInfo bi, TypeMirror excType)
...
 
 			breakpoints[request] = bi;
+		}
+
+		#region ExceptionRequest
+		//VS in Windows doesn't have the concept of a Catchpoint. In the past we did use it anyway, but that leads to all sort of issues and bugs.
+		//The most important difference is that subclasses are never caught, and that Others is a concept that the Catchpoint doesn't support properly.
+		//This region is for Windows, to be able to properly handle exceptions with an exception list. The concept of Other is supported starting with vm version 2.54.
+
+		Dictionary<string, ExceptionEventRequest> exceptionRequests = new Dictionary<string, ExceptionEventRequest>();
```
jstedfast: This should be marked `readonly` and when the debug session ends should be cleared.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -1212,20 +1215,88 @@ void InsertCatchpoint (Catchpoint cp, BreakInfo bi, TypeMirror excType)
...
+		ExceptionEventRequest otherExceptions;
```
jstedfast: This needs to be reset back to `null` when the session ends or it will be out-of-sync if/when the session gets reused.

Mono.Debugging/Mono.Debugging.Client/Catchpoint.cs
```diff
@@ -45,7 +46,14 @@ public Catchpoint (string exceptionName, bool includeSubclasses)
 			this.exceptionName = exceptionName;
 			this.includeSubclasses = includeSubclasses;
 		}
-		
+
+		public Catchpoint (string exceptionName, bool includeSubclasses, bool everythingElse)
+		{
```
jstedfast: Should the other .ctors chain to this one rather than duplicating the initialization logic?

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: I don't think I can mark the collection readonly, since I'm going to be modifying it unless I'm misrepresenting readonly?

Mono.Debugging.Soft/SoftDebuggerSession.cs
joj: Correcting

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -594,6 +594,8 @@ public override void Dispose ()
 
 			sourceFilesDebugInfo.Clear ();
 
+			exceptionRequests.Clear();
+
```
jstedfast: `otherExceptions = null;`

joj: Fixed

PR #260, opened 2019-11-12
Added support for overriding symbol locations

joj:


joj:
@monojenkins backport to d16-5

PR #264, opened 2019-11-25
Only resume VM when suspended.

lateralusX:
Running VM with suspend="n" (for late attach scenarios), raise a number of errors in debugger client. This is due to runtime not being suspended for several events, but debugger client didn't check the suspend policy returned by VM and always tried to do a resume triggering a VMNotSuspendedException for several different event types.

Fix is to check the returned suspend policy (since that indicates if VM is in a suspend state or not) and only resume when suspended.

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -1731,7 +1731,8 @@ namespace Mono.Debugging.Soft
 			}
 
 			try {
-				vm.Resume ();
+				if (es.SuspendPolicy != SuspendPolicy.None)
+					vm.Resume ();
 			} catch (VMNotSuspendedException) {
 				if (type != EventType.VMStart && vm.Version.AtLeast (2, 2))
 					throw;


jstedfast:
I have to revert this because `vm.SuspendPolicy` is *always* `SuspendPolicy.None` in MonoDevelop which means this completely breaks debugging in MonoDevelop.

lateralusX:
OK, any idea what VM version it was using, since looking at the code around this in Mono VM, it should send a SuspendPolicy.None back to the client if the VM has not been resumed, so if the client then does the resume call, you will get VMNotSuspendedException.

lateralusX:
I will do a test with local MonoDevelop build at some point as well to see how it could get into this behavior since it seems to be inconsistent with runtime behavior, especially if you run with suspend=n (that MonoDevelop probably never have done), since that would have lead to a series of exceptions when always trying to resume the runtime without looking at current suspend policy.

jstedfast:
This was with Mono 6.8:

Mono JIT compiler version 6.8.0.73 (2019-10/77147e75266 Tue Dec 10 09:10:17 EST 2019)

lateralusX:
OK, and latest MonoDevelop from master as well?

jstedfast:
Correct, latest monodevelop from master.

How does VM.SuspendPolicy get set? Does that come from the runtime? I wonder if maybe it only gets set if you pass an explicit suspend policy option as an argument to the runtime?

lateralusX:
Its returned from runtime and should reflect how the runtime gets suspended, so when you get none back, it means it is not suspended and doesn't need to be resumed. I tested this under VS debugger having the runtime in both suspend=y (default) and suspend=n (used for late attach) and couldn't observe this scenario, but apparently it happens with MonoDevelop. Lets revert this for now if it causes issues in MonoDevelop and I can research a little more when I get some more bandwidth.

jstedfast:
FWIW, when I discovered this, I was stepping thru `SoftDebuggerSession.HandleEventSet()` and the event type was `EventType.VMStart`.

The repro case I was using was just debugging a standard Console app with a breakpoint set in Main().

BTW, just taking a quick look thru the `VirtualMachine.cs` code - it looks like what we might *actually* want to look at here is the `EventSet.SuspendPolicy` property and not the `VirtualMachine.SuspendPolicy` property.

Does that seem to make sense?

jstedfast:
Never mind, that's exactly what you were checking. There is no VirtualMachine.SuspendPolicy property.

lateralusX:
@jstedfast Looking back a little on this PR (mainly since I'm about to upstream workaround for suspend=n in mono/mono cause by this issue in debugger-libs). So looks like you probably got hit by a separate issue above, looking at this history  https://github.com/mono/mono/pull/19317 seems to be fixing that in case you run with suspend=y. Possible to retry with this PR together with changes in https://github.com/mono/mono/pull/19317 and see if we could get this PR included again?

PR #267, opened 2019-12-04
Use a different lock to manage breakpointStore

sgmunn:
Fixes VSTS 1026591
SigTerm signal in Mono.Debugging.dll!Mono.Debugging.Client.DebuggerSession::get_Breakpoints+9

`slock` is used throughout DebuggerSession, however in this case we only need to guard against
multiple threads assigning an instance of BreakpointStore. Also, since BreakpointStore is thread-safe
we can remove the additional locks around it where we lock on enumeration of `breakpointStore`.

This fixes the case where one thread dispatches an OnExit (for example) which tries communicate
with the debuggee and waits on data from the socket and another thread attempts to clean up breakpoints.

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -254,6 +254,12 @@ namespace Mono.Debugging.Client
 					ownedBreakpointStore = false;
 
 					if (breakpointStore != null) {
+						breakpointStore.BreakEventAdded += OnBreakpointAdded;
+						breakpointStore.BreakEventRemoved += OnBreakpointRemoved;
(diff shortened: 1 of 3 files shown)


sgmunn:
@slluis this does not pass the tests - https://github.com/mono/monodevelop/pull/9431

jstedfast:
The reason this was failing was because of a commit that made it so that `vm.Resume()` did not get called properly after SoftDebuggerSession received a `VMStart` event, so the vm would just hang.

jstedfast:
I've got a new build of https://github.com/mono/monodevelop/pull/9431 going - if that passes the unit tests and DDRITs, then his should be safe to merge.

jstedfast:
Tests are passing

Therzok:
nice!

PR #268, opened 2019-12-04
Separate out calls to OnXxxxx from locking `breakpoints`

sgmunn:
Fixes VSTS 1026593
[FATAL] SigTerm signal in Mono.Debugging.dll!Mono.Debugging.Client.DebuggerSession::GetBreakEventStatus+11

We can deadlock if we allow subclasses to perform arbitrary code inside locks.

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -646,43 +646,54 @@ namespace Mono.Debugging.Client
 
 		bool RemoveBreakEvent (BreakEvent be)
 		{
+			BreakEventInfo binfo = null;
 			lock (breakpoints) {
-				if (breakpoints.TryGetValue (be, out var binfo)) {
-					try {
-						OnRemoveBreakEvent (binfo);
-					} catch (Exception ex) {
-						if (IsConnected)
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -647,42 +647,54 @@ void AddBreakEvent (BreakEvent be)
...
 		}
 		
 		void UpdateBreakEvent (BreakEvent be)
 		{
+			BreakEventInfo binfo;
 			lock (breakpoints) {
-				if (breakpoints.TryGetValue (be, out var binfo))
-					OnUpdateBreakEvent (binfo);
+				breakpoints.TryGetValue (be, out binfo);
+			}
```
jstedfast: We could probably do:

```csharp
lock (breakpoints) {
    if (!breakpoints.TryGetValue (be, out binfo))
        return;
}
```

sgmunn:
tests fail in MonoDevelop

jstedfast:
Just rebased this onto master to get the vm.Resume() fix. Once https://github.com/mono/monodevelop/pull/9432 shows the tests all passing, this should be safe to merge.

jstedfast:
All tests in monodevelop passed.

PR #269, opened 2019-12-05
Invalid Argument exception while setting exception filter.

thaystg:
After the new implementation of exception filters, we need to check the types that are passed by user and the ones that are already in the exception list, and we can receive some wrong information like a exceptionType = "test[" this is not possible to be parsed so we will get and INVALID_ARGUMENT exception, this PR is logging this information and ignoring the error, because this shouldn't stop the debug process.


joj:
@monojenkins backport to d16-4

PR #274, opened 2019-12-05
Get rid of the need for DebuggerSession.set_Breakpoints

jstedfast:
Setting up the breakpoint store in the .ctor allows us
to remove the complexity of set_Breakpoints which isn't
really needed anyway since the breakpoints are only ever
set immediately after creating the session.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1026591/


slluis:
Please add a more respectful and informative description to the PR

PR #276, opened 2019-12-10
Added hooks for gathering telemetry statistics

jstedfast:


sgmunn:
I wonder, would a min and a max / count property be useful? I am thinking that average alone might be a little coarse. at any rate, we can add these or whatever we need to calculate the spread of values if we feel that we need it later.

Therzok:
I think something like [BucketTimings] is better suited here: http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core.Instrumentation/BucketTimings.cs,4c437bca47b07f38

It has an array that uses 2^n <-> 2^(n+1) ranges to bucket specific timings, and then you know how many operations are within a specific duration range.

sgmunn:
actually, I think I misread something. we do have min and max.

sgmunn:
@Therzok can we can do the bucketing in monodevelop, or should we do it here?

Therzok:
@sgmunn if there is a way to get notified of a specific duration in MonoDevelop, it can be done there.

PR #279, opened 2020-01-08
[Mono.Debugging] Fixed logic for timing Step In/Out/Over/etc

jstedfast:


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -1187,6 +1184,12 @@ internal protected void OnTargetEvent (TargetEventArgs args)
 			if (args.Thread != null && args.IsStopEvent)
 				activeThread = args.Thread;
 
+			if (stepTimer != null && (HasExited || args.IsStopEvent)) {
+				stepTimer.Stop (true);
+				stepTimer.Dispose ();
+				stepTimer = null;
+			}
```
mrward: Looking at this now we no longer know if there was an error in the stepping? Before if an exception was thrown the timer.Success flag would not be set to true.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -486,6 +487,12 @@ public void Detach ()
 			}
 		}
 
+		void StartStepTimer (DebuggerStatistics stats)
+		{
+			stepTimer?.Dispose ();
```
mrward: If the stepTimer is disposed here then it was not stopped in the normal way and another timer was started before? Is the timing still useful in this case?

jstedfast: I don't think this can actually happen since the step buttons become deactivated, but I was covering my bases. If we get any telemetry data with a FailedCount, then we know that we are hitting this.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
jstedfast: True, but I doubt it is a very common scenario and if we care about that, we should provide it another way.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
jstedfast: Just to clarify: If the timer gets disposed and the timer was still running, then it will report failure.

jstedfast:
This is the corresponding monodevelop PR: https://github.com/mono/monodevelop/pull/9524

PR #283, opened 2020-01-22
[Mono.Debugging] Refactored code to avoid deadlocks in DebuggerSessio…

jstedfast:
…n when using breakpoints

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1026593/

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -1281,23 +1281,22 @@ namespace Mono.Debugging.Client
 		/// </remarks>
 		internal protected void BindSourceFileBreakpoints (string fullFilePath)
 		{
+			Dictionary<BreakEvent, BreakEventInfo> breakpointsCopy;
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -1338,23 +1337,24 @@ void RetryEventBind (BreakEventInfo binfo)
...
+						toUpdate.Add (bps.Value);
 				}
+			}
 
-				foreach (var be in toUpdate) {
-					breakpoints.Remove (be.BreakEvent);
-					NotifyBreakEventStatusChanged (be.BreakEvent);
-				}
+			foreach (var be in toUpdate) {
+				breakpoints.Remove (be.BreakEvent);
```
mrward: I think we need a lock around this breakpoints.Remove.

jstedfast: Eek, yea, I missed that...

PR #285, opened 2020-01-23
Remove the need to have locking for DebuggerSession.Breakpoints

jstedfast:
Setting up the breakpoint store in the .ctor allows us to remove the complexity of DebuggerSession.set_Breakpoints() which isn't really needed anyway since the breakpoints are only ever set immediately after creating the session.


jstedfast:
Closing this since I'm not sure it's needed

PR #290, opened 2020-02-07
[d16-5] null-protect stepTimer.Stop()/Dispose() in DebuggerSession.OnTargetEv…

jstedfast:
…ent()

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1060768/

Backport of https://github.com/mono/debugger-libs/pull/289

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -1195,10 +1195,12 @@ namespace Mono.Debugging.Client
 			if (args.Thread != null && args.IsStopEvent)
 				activeThread = args.Thread;
 
-			if (stepTimer != null && (HasExited || args.IsStopEvent)) {
-				stepTimer.Stop (true);
-				stepTimer.Dispose ();
-				stepTimer = null;
+			if (HasExited || args.IsStopEvent) {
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -1196,8 +1196,8 @@ internal protected void OnTargetEvent (TargetEventArgs args)
 				activeThread = args.Thread;
 
 			if (stepTimer != null && (HasExited || args.IsStopEvent)) {
-				stepTimer.Stop (true);
-				stepTimer.Dispose ();
+				stepTimer?.Stop (true);
```
Therzok: But there's a null check right above?

Therzok: Capture stepTimer in a local varaible before it's used, and then you can guarantee it's the same one being used.

Therzok: If this code is multi-threaded, atomic swaps should be used:

```csharp
var timer = Interlocked.Exchange (stepTimer, null);
if (timer != null && ..) {
    timer.Stop (true);
}

```

jstedfast: thanks for the tip!

PR #291, opened 2020-02-07
Removed unnecessary locking in DebuggerSession.ActiveThread

jstedfast:
Also added null-protection to SetNextStatement's use of ActiveThread.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1030369/

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -470,9 +470,7 @@ namespace Mono.Debugging.Client
 		/// </remarks>
 		public ThreadInfo ActiveThread {
 			get {
-				lock (slock) {
-					return activeThread;
-				}
+				return activeThread;
 			}
 			set {
 				lock (slock) {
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -470,9 +470,7 @@ public void Detach ()
 		/// </remarks>
 		public ThreadInfo ActiveThread {
 			get {
-				lock (slock) {
-					return activeThread;
-				}
+				return activeThread;
```
Therzok: activeThread is written to outside the lock in [OnTargetEvent](https://github.com/mono/debugger-libs/blob/3c2aeea23797414e86fa225710b217d2fe3aa50e/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs#L1203-L1204)

jstedfast: Yep, I saw that. Are you saying that this lock is needed? or that locking is needed in OnTargetEvent? or both?

Therzok: Oh, it seems like reference assignment is also atomic. Should be good 👍

PR #293, opened 2020-02-11
[UnitTests] Added helper GetAllLocalsSync() method

jstedfast:
This patch is needed by a fix to the VSCodeDebugProtocol backend unit tests

This change effectively just makes sure that all of the Local Variable
objects are fetched synchronously. When a debugger backend's GetAllLocals()
implementation returns an EvalutingGroup dummy variable so that the local
variables can be fetched asynchronously, the new GetAllLocalsSync() method
will evaluate & replace that EvaluatingGroup dummy variable with the real
list of local variables that the unit tests expect to find.


jstedfast:
Merging since the vsmac unit tests easily confirm this fix works ;-)

PR #297, opened 2020-02-14
[Mono.Debugging] Added hooks for gathering usage stats for debugger pads

jstedfast:


Mono.Debugging/Mono.Debugging.Client/UsageCounter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+		public UsageCounter ()
+		{
+		}
+
+		public bool HasUsageStats {
+			get { return stats.Count > 0; }
+		}
+
+		public void IncrementUsage (string action)
```
Therzok: Are there any thread safety concerns?

jstedfast: No, these are all used from the main thread only

PR #299, opened 2020-03-02
[debugger] Implementing step through multithreaded code.

thaystg:
I've changed the variable the_ss_req to an array, in this array I saved all the single steps requisitions received and whenever the debugger stops in a possible singles step I get the right requisition in the array and answer it.
It is necessary to change debugger-libs together, because we can send a single step for thread A and receive an answer of thread B, so we shouldn't cancel the single step requisition for thread A, because it wasn't answered yet.
If the change in debugger-libs is not synchronised together if mono we will not have any side effect, the multithreaded step just don't work as nowadays.

Fixes #14456
Relates to https://github.com/mono/mono/pull/19103

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/EventRequest.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/EventRequest.cs
@@ -34,6 +34,11 @@ namespace Mono.Debugger.Soft
 			}
 		}
 
+		public int getId ()
+		{
+			return id;
+		}
+
 		public EventType EventType {
 			get {
 				return etype;
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/EventRequest.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/EventRequest.cs
@@ -34,7 +34,7 @@ namespace Mono.Debugger.Soft
 			}
 		}
 
-		public int getId ()
+		public int GetId ()
 		{
 			return id;
 		}
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
@@ -613,9 +613,9 @@ namespace Mono.Debugger.Soft
 
 		internal EventRequest GetRequest (int id) {
 			lock (requests_lock) {
-				if (requests.ContainsKey(id))
-					return requests [id];
-				return null;
+				EventRequest obj;
+				requests.TryGetValue (id, out obj);
+				return obj;
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -1947,10 +1947,11 @@ void HandleBreakEventSet (Event[] es, bool dequeuing)
 						}
 						
 						if (hasBreakInfo) {
-							if (currentStepRequest != null &&
-							    currentStepRequest.Depth != StepDepth.Out &&
-							    binfo.Location.ILOffset == currentAddress && 
-							    e.Thread.Id == currentStepRequest.Thread.Id &&
+							if (currentRequest != null &&
+								currentRequest is StepEventRequest &&
+								(currentRequest as StepEventRequest).Depth != StepDepth.Out &&
```
jstedfast: This could probably be expressed like this:

```csharp
if ((currentRequest as StepEventRequest currentStepRequest) && currentStepRequest.Depth != StepDepth.Out &&
```

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -1969,27 +1970,30 @@ void HandleBreakEventSet (Event[] es, bool dequeuing)
...
+			}
+
+
 			if (redoCurrentStep) {
-				StepDepth depth = currentStepRequest.Depth;
-				StepSize size = currentStepRequest.Size;
-				
+				StepDepth depth = (currentRequest as StepEventRequest).Depth;
+				StepSize size = (currentRequest as StepEventRequest).Size;
+
```
jstedfast: These should probably be regular casts instead of as-casts, unless they can be a request type other than StepEventRequest, in which case this code would need to handle the as-cast resulting in null.

thaystg: Done

Mono.Debugging.Soft/SoftDebuggerSession.cs
thaystg: Done

Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
```diff
@@ -613,7 +613,9 @@ class MirrorCache<T> {
 
 		internal EventRequest GetRequest (int id) {
 			lock (requests_lock) {
-				return requests [id];
+				if (requests.ContainsKey(id))
```
slluis: nit: use `TryGetValue` to avoid double access

thaystg: Done! :)

Mono.Debugger.Soft/Mono.Debugger.Soft/EventRequest.cs
```diff
@@ -34,6 +34,11 @@ public abstract class EventRequest {
 			}
 		}
 
+		public int getId ()
```
nerzhulart: Just codestyle: the property must be called with capital first letter

thaystg: Done.

PR #300, opened 2020-03-04
[debugger] NRE when 2 threads try to call GetThreads at same time

thaystg:
As analysed by @jstedfast and me the only way to get this exception with this callstack below is 2 threads call GetThreads at same time, so while one of them is filling the items in the array, the other one is already trying to use the array with null items.

```
System.NullReferenceException: Object reference not set to an instance of an object
  at Mono.Debugging.Soft.SoftDebuggerSession.GetId (Mono.Debugger.Soft.ThreadMirror thread) [0x00000] in /Users/runner/runners/2.164.6/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:3227
  at Mono.Debugging.Soft.SoftDebuggerSession.OnGetThreads (System.Int64 processId) [0x0002c] in /Users/runner/runners/2.164.6/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:879
  at Mono.Debugging.Soft.SoftDebuggerSession.GetThread (Mono.Debugging.Client.ProcessInfo process, Mono.Debugger.Soft.ThreadMirror thread) [0x00008] in /Users/runner/runners/2.164.6/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:902
  at Mono.Debugging.Soft.SoftDebuggerSession.HandleBreakEventSet (Mono.Debugger.Soft.Event[] es, System.Boolean dequeuing) [0x00501] in /Users/runner/runners/2.164.6/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:2038
  at Mono.Debugging.Soft.SoftDebuggerSession.HandleEventSet (Mono.Debugger.Soft.EventSet es) [0x0005b] in /Users/runner/runners/2.164.6/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:1687
  at Mono.Debugging.Soft.SoftDebuggerSession.EventHandler () [0x0005d] in /Users/runner/runners/2.164.6/work/1/s/monodevelop/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:1625
```

PR on mono/mono -> https://github.com/mono/mono/pull/19122

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1070384/

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
@@ -197,6 +197,7 @@ namespace Mono.Debugger.Soft
 				//if (threadCache != threads) {//While fetching threads threadCache was invalidated(thread was created/destoyed)
 				//	return GetThreads ();
 				//}
+				Thread.MemoryBarrier ();
 				threadCache = threads;
 				return threads;
 			} else {
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
@@ -177,30 +177,32 @@ namespace Mono.Debugger.Soft
 		}
 
 		public IList<ThreadMirror> GetThreads () {
-			var threads = threadCache;
-			if (threads == null) {
-				long[] ids = null;
-				var fetchingEvent = new ManualResetEvent (false);
-				vm.conn.VM_GetThreads ((threadsIds) => {
-					ids = threadsIds;
-					threadCache = threads = new ThreadMirror [threadsIds.Length];
-					fetchingEvent.Set ();
-				});
-				if (WaitHandle.WaitAny (new []{ vm.conn.DisconnectedEvent, fetchingEvent }) == 0) {
-					throw new VMDisconnectedException ();
+			lock (threadCacheLocker) {
+				var threads = threadCache;
+				if (threads == null) {
+					long[] ids = null;
+					var fetchingEvent = new ManualResetEvent (false);
+					vm.conn.VM_GetThreads ((threadsIds) => {
+						ids = threadsIds;
+						threadCache = threads = new ThreadMirror [threadsIds.Length];
+						fetchingEvent.Set ();
+					});
+					if (WaitHandle.WaitAny (new []{ vm.conn.DisconnectedEvent, fetchingEvent }) == 0) {
+						throw new VMDisconnectedException ();
+					}
+					for (int i = 0; i < ids.Length; ++i)
+						threads [i] = GetThread (ids [i]);
+					//Uncomment lines below if you want to re-fetch threads if new threads were started/stopped while
+					//featching threads... This is probably more correct but might cause deadlock of this method if runtime
+					//is starting/stopping threads nonstop, need way to prevent this(counting number of recursions?)
+					//possiblity before uncommenting
+					//if (threadCache != threads) {//While fetching threads threadCache was invalidated(thread was created/destoyed)
+					//	return GetThreads ();
+					//}
+					return threads;
+				} else {
+					return threads;
 				}
-				for (int i = 0; i < ids.Length; ++i)
-					threads [i] = GetThread (ids [i]);
-				//Uncomment lines below if you want to re-fetch threads if new threads were started/stopped while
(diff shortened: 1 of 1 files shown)


lambdageek:
@jstedfast it looks like there used to be a lock already a few years ago, but then we lost it during some refactoring?  are there other things that need to be protected by that lock?

jstedfast:
@lambdageek I'm the last person to ask about that, I wouldn't have any idea.

lambdageek:
Hmm.... looks like the lock around the cache was dropped in https://github.com/mono/mono/commit/948689360ba96ab29b109ae353e678ef633e2675  with a comment about a deadlock.

I wonder if we're reintroducing it again

Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
```diff
@@ -177,30 +177,32 @@ internal void AddThreadToInvalidateList (ThreadMirror threadMirror)
...
-				if (WaitHandle.WaitAny (new []{ vm.conn.DisconnectedEvent, fetchingEvent }) == 0) {
-					throw new VMDisconnectedException ();
+			lock (threadCacheLocker) {
+				var threads = threadCache;
+				if (threads == null) {
+					long[] ids = null;
+					var fetchingEvent = new ManualResetEvent (false);
+					vm.conn.VM_GetThreads ((threadsIds) => {
+						ids = threadsIds;
+						threadCache = threads = new ThreadMirror [threadsIds.Length];
```
DavidKarlas: Maybe this can be fixed without locking by chaning `threadCache = threads = new ThreadMirror [threadsIds.Length];` to `threads = new ThreadMirror [threadsIds.Length];`

Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
```diff
@@ -177,30 +177,32 @@ internal void AddThreadToInvalidateList (ThreadMirror threadMirror)
...
+						fetchingEvent.Set ();
+					});
+					if (WaitHandle.WaitAny (new []{ vm.conn.DisconnectedEvent, fetchingEvent }) == 0) {
+						throw new VMDisconnectedException ();
+					}
+					for (int i = 0; i < ids.Length; ++i)
+						threads [i] = GetThread (ids [i]);
+					//Uncomment lines below if you want to re-fetch threads if new threads were started/stopped while
```
DavidKarlas: And after this loop, set `threadCache=threads;`

thaystg: Yes yes, this is another possible solution that i discussed with @jstedfast, but we were afraid to call much more times the VM_GetThreads than it's really necessary. 
But with @lambdageek analysis, I think we prefer it than have a deadlock again.
I will do this.

Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
```diff
@@ -198,6 +197,7 @@ internal void AddThreadToInvalidateList (ThreadMirror threadMirror)
 				//if (threadCache != threads) {//While fetching threads threadCache was invalidated(thread was created/destoyed)
 				//	return GetThreads ();
 				//}
+				threadCache = threads;
```
lambdageek: I think technically you need a call to `Thread.MemoryBarrier()` before this assignment so that all the writes to the `threads` array are seen by other threads.  Otherwise they may still see null.  (I guess we only ever run this code on ARM on Mono's CI, but still...)

thaystg: Done

DavidKarlas:
After giving some more thought to this... most correct way to fix this is to change return type to `Task<IList<ThreadMirror>>` which would allow us to create TaskCompletionSource and following calls would get same Task as 1st call...
Ofc this means whole callstack will need to be asynced...

lambdageek:
> change return type to `Task<IList<ThreadMirror>>`

That sounds like a good follow-up task as part of a larger effort to async-ify more of debugger-libs

PR #309, opened 2020-04-03
Fix side-effect of multithreaded ss

thaystg:
It was cancelling a request even if it was a breakpoint request and not only when it was a step request.
Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1092604

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -1980,17 +1980,14 @@ namespace Mono.Debugging.Soft
 				currentRequest = es[0].Request;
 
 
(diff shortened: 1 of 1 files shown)


jstedfast:
When trying to step over a line with a breakpoint, I get:

System.NotSupportedException: This request is not supported by the protocol version implemented by the debuggee.
  at Mono.Debugging.Soft.SoftDebuggerSession.Step (Mono.Debugger.Soft.StepDepth depth, Mono.Debugger.Soft.StepSize size) [0x000b1] in /Users/fejj/src/vsmac/main/external/debugger-libs/Mono.Debugging.Soft/SoftDebuggerSession.cs:1598 

Hat said, this fix *does* seem to make it so that I can "Continue" and have breakpoints be hit again.

PR #306, opened 2020-03-31
[Mono.Debugging.Tests] Disabled Sdb tests

VladyslavKostin:
Specification
Disable test due to issue
Bug 1090662: Timeout while waiting for initial breakpoint SdbAdvancedEvaluationAllowTargetInvokesTests (Mono.Debugging.Tests.Soft.SdbAdvancedEvaluationAllowTargetInvokesTests)
https://work.azdo.io/1090662

PR #312, opened 2020-04-09
[Mono.Debugging.Soft] Simplified breakpoint resolving logic

jstedfast:
The old logic was trying to be so smart, it smartness-overloaded
into Smartness.MinValue.

In all seriousness, though, now that the IDE intelligently resolves
breakpoints to the correct line, we don't need to be "smart".

This code was only needed 10 years ago when the IDE would put
breakpoints on blank lines and so the debugger would need to
do insanely fuzzy matching. So fuzzy it needed glasses, apparently.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1094719/

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -38,51 +38,27 @@ using System.Net.Sockets;
 using System.Globalization;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
-
-using Mono.CompilerServices.SymbolWriter;
-using Mono.Debugging.Client;
-using Mono.Debugger.Soft;
-using Mono.Debugging.Evaluation;
-using MDB = Mono.Debugger.Soft;
-using System.Security.Cryptography;
-using Mono.Cecil.Cil;
-using Newtonsoft.Json;
 using System.Text.RegularExpressions;
+
+using Newtonsoft.Json;
+
+using Mono.Debugger.Soft;
+
+using Mono.Debugging.Client;
+using Mono.Debugging.Evaluation;
+
+using StackFrame = Mono.Debugger.Soft.StackFrame;
 
 namespace Mono.Debugging.Soft
 {
 	public class SoftDebuggerSession : DebuggerSession
 	{
-		class SourceFileDebugInfo
-		{
-			public string FullFilePath { get; set; }
-			public int FileID { get; set; }
-			public byte [] Hash { get; set; }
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
Difficult to eyeball it from just reading, but I think it makes sense overall.

A few corner cases to think about: 

1. multiple methods covering the same exact span (lambda body inside a method body) - choose the method with the innermost body
2. multiple sequence points on the same line (e.g. `for (int i = 0; i < n; i++)`)
3. multiple methods on the same line ` { get; set; } `

Would it be easy to add a few tests for these scenarios? Or do we not have the proper test infrastructure in this area?

If not, perhaps it's worth it to just create a simple console app with the above scenarios and try it out? Remember to try both debug and release. Release usually gives you less information and you can place breakpoints in less places.

DavidKarlas:
I'm also not convinced breakpoints in `async anonymous` methods will work when placed before debugging starts.

DavidKarlas:
> This is a good chance to add a comprehensive test suite covering breakpoint resolving

Good first step would be re-enabling existing ones: https://github.com/mono/debugger-libs/blob/master/UnitTests/Mono.Debugging.Tests/Shared/BreakpointsAndSteppingTests.cs were disabled here: https://github.com/mono/debugger-libs/pull/306

jstedfast:
@DavidKarlas 

> I'm also not convinced breakpoints in async anonymous methods will work when placed before debugging starts.

You were right, it did not work.

The code works now, though.

jstedfast:
@KirillOsenkov 

> A few corner cases to think about:
> 1. multiple methods covering the same exact span (lambda body inside a method body) - choose the method with the innermost body

This is handled at a higher level by the IDE's BreakpointSpanResolver.

> 2. multiple sequence points on the same line (e.g. for (int i = 0; i < n; i++))

Yep, this is fixed now and there exist unit tests for this (just got all the unit tests passing again).

> 3. multiple methods on the same line { get; set; }

Also works thanks to the IDE's BreakpointSpanResolver and code changes since original PR.

slluis:
@jstedfast So what happened with my request of adding unit tests?

Therzok:
@jstedfast we had tests for that, and they were re-enabled.

slluis:
Re-enabled where?

jstedfast:
@slluis: https://github.com/mono/debugger-libs/blob/master/UnitTests/Mono.Debugging.Tests/Shared/BreakpointsAndSteppingTests.cs#L586

PR #318, opened 2020-04-20
[Mono.Debugging] Added SackFrame.GetFullStackFrameText/Async()

jstedfast:
Part of a larger fix for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1083539/


Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
```diff
@@ -122,63 +123,98 @@ internal void Attach (DebuggerSession debugSession)
...
-										session.OnDebuggerOutput (true, "Timeout evaluating parameters for StackFrame.");
-								args [n].ValueChanged -= updated;
-							}
-							if (n > 0)
-								methodNameBuilder.Append (", ");
-							if (options.StackFrameFormat.ParameterTypes) {
-								methodNameBuilder.Append (args [n].TypeName);
-								if (options.StackFrameFormat.ParameterNames)
-									methodNameBuilder.Append (" ");
+		public virtual string GetFullStackFrameText (EvaluationOptions options)
```
Therzok, line 136: Just pass in a cancellationtoken in EvaluationOptions and use that?

Therzok: This would avoid hardcoding a 2s timer here, which could probably not be enough. Also, shouldn't we allow only the async versino to be override-able?

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
```diff
@@ -122,63 +123,98 @@ internal void Attach (DebuggerSession debugSession)
...
+								tcs.TrySetResult (true);
+							};
+							args[n].ValueChanged += updated;
+							try {
+								using (var registration = cancellationToken.Register (() => tcs.TrySetCanceled ())) {
+									if (args[n].IsEvaluating) {
+										if (doAsync) {
+											await tcs.Task.ConfigureAwait (false);
+										} else {
+											tcs.Task.Wait (cancellationToken);
```
Therzok, line 189: Should we also be using the evaluation timeout here and above by linking a new token source with the one passed in?

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
jstedfast: 2 good points.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
jstedfast: Hmmm, probably.

I also discovered that there's a way to cancel outstanding requests (theoretically if the backend supports it, so I need to check the protocolClient's capabilities).

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
jstedfast: FWIW, the old code used a 2s timer which is where I got that from.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
jstedfast: > Should we also be using the evaluation timeout here and above by linking a new token source with the one passed in?

Actually, you know what? Nah.

Why should we cancel it if it takes longer than a timeout when our UI layer can cancel a token?

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
Therzok: Should we also drop the possibility to override the non-async version? Otherwise we could end up with 2 different implementations based on the API you call.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
```diff
@@ -122,63 +128,110 @@ internal void Attach (DebuggerSession debugSession)
...
-									methodNameBuilder.Append (" ");
-							}
-							if (options.StackFrameFormat.ParameterNames)
-								methodNameBuilder.Append (args [n].Name);
-							if (options.StackFrameFormat.ParameterValues) {
-								if (options.StackFrameFormat.ParameterTypes || options.StackFrameFormat.ParameterNames)
-									methodNameBuilder.Append (" = ");
-								var val = args [n].Value ?? "";
-								methodNameBuilder.Append (val.Replace ("\r\n", " ").Replace ("\n", " "));
+		public virtual string GetFullStackFrameText (EvaluationOptions options)
```
Therzok, line 141:
```suggestion
		public string GetFullStackFrameText (EvaluationOptions options)
```

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
jstedfast: Well, the .NET Core backend overrides the non-async version, so it's still needed.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
jstedfast: Can't because the DotNetCore backend overrides this.

jstedfast:
@slluis the DotNetCore implementation overrides it because t has the vscode backend format the stackframe string.

PR #320, opened 2020-04-29
[Mono.Debugger.Soft] Fixed NRE in MethodMirror.FullName when param_in…

jstedfast:
…fo is not yet cached

If param_info has not yet been fetched, then the FullName property
will throw a NRE when accessed.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1112185/


jstedfast:
Merging this now that it was also accepted into Mono.

PR #322, opened 2020-05-01
[Mono.Debugging.Soft] Invalidate cached local variable values

jstedfast:
This is an attempt to fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1109965/

Unfortunately, while this *does* improve things, it doesn't fix the
issue entirely.

### DIFF
--- a/Mono.Debugging.Soft/LocalVariableBatch.cs
+++ b/Mono.Debugging.Soft/LocalVariableBatch.cs
@@ -32,22 +32,23 @@ namespace Mono.Debugging.Soft
 	public class LocalVariableBatch
 	{
 		readonly LocalVariable[] variables;
-		readonly StackFrame frame;
 		Value[] values;
+		int version;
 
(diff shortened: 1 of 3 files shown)


jstedfast:
This fix obsoletes https://github.com/mono/debugger-libs/pull/321

mrward:
How strict are we are not breaking the public API for debugger-libs?

jstedfast:
@mrward that's a good question. I don't know... Maybe @slluis can comment?

jstedfast:
Before breaking API, we should probably at least notify the Unity and XamarinVS teams since I'm not sure if they would be affected by this API break or not.

/cc @jbevain for Unity ?
/cc @joj from XamarinVS ?

DavidKarlas:
I check and neither Unity nor XamarinVS depends on this API... Its unlikely anyone depends on it...

netonjm:
> I check and neither Unity nor XamarinVS depends on this API... Its unlikely anyone depends on it...

Xamarin.IoT uses this library :-)

PR #326, opened 2020-07-13
Since this is public API we can't add more abstract methods

DavidKarlas:
Hence add virtual so we don't break compatibility...


GouriKumari:
@monojenkins backport d16-8

PR #329, opened 2020-09-01
Added LocalVariableRendering time and more information to timer Start+Tracing

DavidKarlas:
This allows for real time and precise tracing of debugger times

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
@@ -136,7 +136,7 @@ namespace Mono.Debugging.Client
 			this.stats = stats;
 			if (stats != null && traceListener != null) {
 				traceId = Interlocked.Increment (ref traceSeq);
-				traceListener?.TraceEvent (null, stats.Name, TraceEventType.Start, traceId, name);
+				traceListener.TraceEvent (null, stats.Name, TraceEventType.Start, traceId, name);
 			}
 		}
 
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
@@ -27,6 +27,8 @@
 using System;
 using System.Diagnostics;
 using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 
 namespace Mono.Debugging.Client
 {
@@ -38,6 +40,11 @@ namespace Mono.Debugging.Client
 		readonly object mutex = new object ();
 		TimeSpan minTime = TimeSpan.MaxValue;
 		TimeSpan maxTime, totalTime;
+
+		public DebuggerStatistics(string name)
+		{
+			Name = name;
+		}
 
 		public double MaxTime {
 			get { return maxTime.TotalMilliseconds; }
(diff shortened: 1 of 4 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
```diff
@@ -39,6 +41,11 @@ public class DebuggerStatistics
 		TimeSpan minTime = TimeSpan.MaxValue;
 		TimeSpan maxTime, totalTime;
 
+		public DebuggerStatistics(string name)
+		{
+			Name = name;
+		}
```
mrward, line 47: API break? Should we also add the DebuggerStatistics method which takes no parameter?

Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
```diff
@@ -59,12 +66,11 @@ public class DebuggerStatistics
 		public int TimingsCount { get; private set; }
 
 		public int FailureCount { get; private set; }
+		public string Name { get; }
 
-		public DebuggerTimer StartTimer ()
+		public DebuggerTimer StartTimer (string name)
 		{
```
mrward, line 72: API break?

Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
```diff
@@ -136,19 +150,14 @@ public DebuggerTimer (DebuggerStatistics stats)
 			get { return stopwatch.Elapsed; }
 		}
 
-		public void Start ()
```
mrward, line 139: Another API break.

Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
```diff
@@ -120,10 +126,18 @@ public class DebuggerTimer : IDisposable
 		readonly DebuggerStatistics stats;
 		readonly Stopwatch stopwatch;
 
-		public DebuggerTimer (DebuggerStatistics stats)
+		static readonly TraceListener traceListener = new TraceSource (nameof (DebuggerTimer)).Listeners.OfType<TraceListener> ().FirstOrDefault (l => l.Name == nameof (DebuggerTimer));
```
mrward, line 129: Can we comment how to use this? I am assuming you need to setup a trace listener in VS Mac?

Mono.Debugging/Mono.Debugging.Client/DebuggerStatistics.cs
```diff
@@ -120,10 +126,18 @@ public class DebuggerTimer : IDisposable
...
+		int traceId;
+
+		public DebuggerTimer (DebuggerStatistics stats, string name)
 		{
-			stopwatch = new Stopwatch ();
+			stopwatch = Stopwatch.StartNew ();
 			this.stats = stats;
+			if (stats != null && traceListener != null) {
+				traceId = Interlocked.Increment (ref traceSeq);
+				traceListener?.TraceEvent (null, stats.Name, TraceEventType.Start, traceId, name);
```
Therzok, line 139: Already null checked.

```suggestion
				traceListener.TraceEvent (null, stats.Name, TraceEventType.Start, traceId, name);
```

Therzok (approved):

mrward (commented):
Code looks good to me. Do we have a problem with the public API changing? If not I can approve :)

mrward (approved):
Well, only VS Mac seems to use DebuggerStatistics. I suspect nobody else is using it.

DavidKarlas:
About API break, yes, only VSMac is using it and for anyone in future I want to force them to use names for things... Hence removing old overload...
About how to use this, I will post in VSMac PR...

PR #336, opened 2020-11-17
Make fullModuleName protected

KirillOsenkov:
This is so we can set it from a derived class such as VSCodeStackFrame

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
@@ -20,7 +20,6 @@ namespace Mono.Debugging.Client
 		bool isExternalCode;
 		bool isDebuggerHidden;
 		bool hasDebugInfo;
-		protected string fullModuleName;
 		string fullTypeName;
 		
 		[NonSerialized]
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
```diff
@@ -20,7 +20,7 @@ public class StackFrame
 		bool isExternalCode;
 		bool isDebuggerHidden;
 		bool hasDebugInfo;
-		string fullModuleName;
+		protected string fullModuleName;
```
Therzok, line 23: This should use proper naming style conventions.

```suggestion
		protected string FullModuleName;
```

slluis: Better use a protected property instead of a field.

KirillOsenkov: I wanted to minimize changes, also FullModuleName is already a property, so it would mean making a backing field of a property itself a property. We can't use upper case because it would conflict with the FullModuleName property.

Where in the naming conventions do we have specified that protected fields need to start with an uppercase?

KirillOsenkov: Decided to go with an auto-property.

slluis (approved):

PR #337, opened 2020-11-18
Version check in AssemblyMirror.GetPdbBlob

ivorne:
The current version check throws an exception. But the only function that calls this - SoftDebuggerSession.GetPdbData - does a null check. So returning null here instead of throwing an exception will make this fail gracefully instead of interrupting the whole debugging session.

This is specifically useful in situations, when we are debugging an older mono runtime and we only have MDBs and no PDBs.


nosami:
Looks good to me. Sorry for the delay

nosami:
Looks good to me. Sorry for the delay

PR #341, opened 2021-02-10
Fix assembly version after converting projects to SDK style

mrward:
SDK style projects set the assembly version to 1.0.0.0 by
default. Classic projects use 0.0.0.0 as the assembly version.
Setting the version to 0.0.0.0 to avoid compatibility warnings.


jpobst (approved):
Whoops!  Thanks for the fix!

mrward:
VS Mac's binary compat check seems happy now with this change - https://github.com/xamarin/vsmac/pull/3221

Merging this.

PR #342, opened 2021-03-03
Add Module.ApplyChanges

lambdageek:
- [Mono.Debugger.Soft] add Module.ApplyChanges
- [Mono.Debugging.Soft] add SoftDebuggerSession.ApplyChanges

Related mono/mono Mono.Debugger.Soft PR: https://github.com/mono/mono/pull/20889
Related dotnet/runtime PR: https://github.com/dotnet/runtime/pull/49043

Contributes to https://github.com/dotnet/runtime/issues/44806

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -567,6 +567,7 @@ namespace Mono.Debugger.Soft
 
 		enum CmdModule {
 			GET_INFO = 1,
+			APPLY_CHANGES = 2,
 		}
 
 		enum CmdMethod {
@@ -2355,6 +2356,11 @@ namespace Mono.Debugger.Soft
 			if (Version.AtLeast (2, 48))
 				info.SourceLink = r.ReadString ();
 			return info;
+		}
+
+
+		internal void Module_ApplyChanges (long id, long dmeta_id, long dil_id, long dpdb_id) {
+			SendReceive (CommandSet.MODULE, (int)CmdModule.APPLY_CHANGES, new PacketWriter().WriteId (id).WriteId (dmeta_id).WriteId (dil_id).WriteId (dpdb_id));
 		}
 
 		/*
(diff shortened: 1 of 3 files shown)


lambdageek:
@thaystg @joj

thaystg (approved):

joj (approved):

AlexeiNaabal:
@lambdageek Could you provide some guidance about how to get the applying ModuleMirror? 

The way I test hot reload is by 1. send the delta to a debug session frontend implement, 2. then pass them to ApplyChanges method with a class member which is a SoftDebuggerSession instance. This workflow is not triggered by debug event, I send the delta through named pipe.

Not finding public ways to get ModuleMirror, currently I add a custom method to SoftDebuggerSession to search current loaded AssemblyMirror for the applying one (with the assembly name that I send through named pipe). 

Though this workflow worked in my test case, I'm not sure if it is the right way to retrieve the ModuleMirror, since I have to modify the debugger-lib source codes.

lambdageek:
@AlexeiNaabal  I'm not a debugger-libs expert. However, I think you get a ModuleMirror by calling `ManifestModule` on an `AssemblyMirror`.  You get `AssemblyMirror`s by collecting them from `AssemblyLoadEvent`s, or by calling `GetAssemblies` on an `AppDomainMirror`.  

So something like `vm.RootDomain.GetAssemblies().Where(am => am.GetName() == someName).First().ManifestModule`

AlexeiNaabal:
> @AlexeiNaabal I'm not a debugger-libs expert. However, I think you get a ModuleMirror by calling `ManifestModule` on an `AssemblyMirror`. You get `AssemblyMirror`s by collecting them from `AssemblyLoadEvent`s, or by calling `GetAssemblies` on an `AppDomainMirror`.
> 
> So something like `vm.RootDomain.GetAssemblies().Where(am => am.GetName() == someName).First().ManifestModule`

Thanks for your guidance, I found what I've missed immediately :)

In fact I can at least access the `vm.RootDomain` within my debugger front-end (which is for delivering the IDE's debug event to the debugger-lib), so I just need to move this `GetAssemblies().Where` process to my front-end, without touching the source code of 3rd party's library :)

PR #343, opened 2021-05-20
Ported evaluator to roslyn.

mkrueger:
Started to move the evaluator code to roslyn. Completely untested at that point. Do not use - but feel free to add/work on stuff.

NRefactory is unsupported and completely out of date. This is the last part in our IDE which uses it - it should get replaced.

How about using a 3rd party C# scripting evaluator for that? Unfortunately Microsoft.CodeAnalysis.CSharp.Scripting isn't enough for our use case - however it may be worth extending support.

Esp. resolving of identifiers is missing in the public API - if added it may do the job.

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
@@ -483,16 +483,14 @@ namespace Mono.Debugging.Client
 			if (string.IsNullOrEmpty(path))
 				return path;
 
-			string result;
-			try
-			{
-				result = realpath(path, IntPtr.Zero) ?? Path.GetFullPath(path);
+			string result = null;
+			try {
+				result = realpath(path, IntPtr.Zero);
 			}
-			catch
-			{
-				result = Path.GetFullPath(path);
+			catch {
 			}
-			return result;
+
+			return string.IsNullOrEmpty(result) ? Path.GetFullPath(path) : result;
 		}
 
 		public static bool FileNameEquals (string file1, string file2)
--- a/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
@@ -483,10 +483,10 @@ namespace Mono.Debugging.Client
 			if (string.IsNullOrEmpty(path))
 				return path;
 
-			string result = null;
+			string result;
 			try
 			{
-				result = realpath(path, IntPtr.Zero);
+				result = realpath(path, IntPtr.Zero) ?? Path.GetFullPath(path);
 			}
 			catch
 			{
--- a/Mono.Debugging/Mono.Debugging.Evaluation/LambdaBodyOutputVisitor.cs
+++ b/Mono.Debugging/Mono.Debugging.Evaluation/LambdaBodyOutputVisitor.cs
@@ -2,10 +2,10 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Mono.Debugging.Client;
-
-using ICSharpCode.NRefactory.CSharp;
 
 namespace Mono.Debugging.Evaluation
 {
@@ -16,7 +16,7 @@ namespace Mono.Debugging.Evaluation
 	//   identifer.
 	// - Collect variable references for outside the lambda
 	//   (local variables/properties...)
-	public class LambdaBodyOutputVisitor : CSharpOutputVisitor
+	public class LambdaBodyOutputVisitor : CSharpSyntaxWalker
 	{
 		readonly Dictionary<string, ValueReference> userVariables;
 		readonly EvaluationContext ctx;
@@ -25,7 +25,7 @@ namespace Mono.Debugging.Evaluation
 		List<string> definedIdentifier;
 		int gensymCount;
 
-		public LambdaBodyOutputVisitor (EvaluationContext ctx, Dictionary<string, ValueReference> userVariables, TextWriter writer) : base (writer, FormattingOptionsFactory.CreateMono ())
+		public LambdaBodyOutputVisitor (EvaluationContext ctx, Dictionary<string, ValueReference> userVariables, TextWriter writer)
 		{
 			this.ctx = ctx;
 			this.userVariables = userVariables;
@@ -33,12 +33,21 @@ namespace Mono.Debugging.Evaluation
 			this.definedIdentifier = new List<string> ();
 		}
 
+		void WriteKeyword (string keyword)
+		{
+
+		}
+		void WriteIdentifier (string keyword)
+		{
+
+		}
+
 		public Tuple<string, object>[] GetLocalValues ()
 		{
 			var locals = new Tuple<string, object>[localValues.Count];
 			int n = 0;
-			foreach(var localv in localValues.Values) {
-				locals [n] = localv;
+			foreach (var localv in localValues.Values) {
+				locals[n] = localv;
 				n++;
 			}
 			return locals;
@@ -54,7 +63,7 @@ namespace Mono.Debugging.Evaluation
 			return new NotSupportedExpressionException ();
 		}
 
-		static Exception EvaluationError (string message, params object [] args)
+		static Exception EvaluationError (string message, params object[] args)
 		{
 			return new EvaluatorException (message, args);
 		}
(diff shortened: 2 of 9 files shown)


Mono.Debugging/Mono.Debugging.Evaluation/LambdaBodyOutputVisitor.cs
```diff
@@ -296,527 +250,51 @@ public override void VisitInvocationExpression (InvocationExpression invocationE
...
-		{
-		}*/
 
-		public override void VisitNamedArgumentExpression (NamedArgumentExpression namedArgumentExpression)
+		public override void VisitSimpleLambdaExpression (SimpleLambdaExpressionSyntax node)
 		{
-			throw NotSupportedToConsistency ();
-		}
+			if (node.Parameter.Modifiers.Count > 0)
+				throw NotSupported ();
```
Therzok, line 269: you can have static and async lambdas.

Therzok: And static lambdas, at least, shouldn't impact anything.

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs
```diff
@@ -432,47 +434,49 @@ ValueReference EvaluateBinaryOperatorExpression (BinaryOperatorType op, ValueRef
...
+		static string ResolveType (EvaluationContext ctx, MemberAccessExpressionSyntax mre, List<object> args)
 		{
 			string parent, name;
 
-			if (mre.Target is MemberReferenceExpression) {
-				parent = ResolveType (ctx, (MemberReferenceExpression) mre.Target, args);
-			} else if (mre.Target is TypeReferenceExpression) {
+			if (mre.Expression is MemberAccessExpressionSyntax mae) {
+				parent = ResolveType (ctx, mae, args);
+			} else /* TODO?
```
Therzok, line 469: All the type resolving is gone?

Therzok (changes requested):
The type resolution is gone and lambda support has some constraints which it shouldn't have?

srxqds:
why this pr still not got merged?

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs
mkrueger: TypeReferenceExpression is gone.

mkrueger:
It needs to be tested. Someone with more knowlegde about the debugger needs to take it from here.

mkrueger:
Who says Type resulution is gone? TypeReferenceExpression is no part of the Roslyn AST. NRefactory and Roslyn ASTs differ. Has nothing to do with resolution.

srxqds:
> It needs to be tested. Someone with more knowlegde about the debugger needs to take it from here.

so, hope can be tested and merged

nosami:
This on the radar for 17.5. I'll get this tested

srxqds:
> This on the radar for 17.5. I'll get this tested

thank you for your work

Mono.Debugging/Mono.Debugging.csproj
```diff
@@ -16,14 +16,10 @@
...
 
-  <ItemGroup>
-    <ProjectReference Include="..\..\nrefactory\ICSharpCode.NRefactory\ICSharpCode.NRefactory.csproj" />
-    <ProjectReference Include="..\..\nrefactory\ICSharpCode.NRefactory.CSharp\ICSharpCode.NRefactory.CSharp.csproj" />
-  </ItemGroup>
-
   <ItemGroup>
     <PackageReference Include="System.Buffers" Version="4.5.0" />
     <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.8.0" />
```
sandyarmstrong, line 22: May as well bump some of these while you're here. Immutable can go to 6.0.0. Roslyn can go to 4.4.0-3.22505.14.

Also you'll want to use msbuild properties for these versions so they can be overridden by the vsmac build. Just like here:

https://github.com/mono/debugger-libs/blob/a832ea7e752c85e0409eccfdc1a2650261be36df/Mono.Debugging.Soft/Mono.Debugging.Soft.csproj#L21-L28

You can use `$(NuGetVersionSystemCollectionsImmutable)` and `$(NuGetVersionRoslyn)`.

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExtensions.cs
```diff
@@ -69,50 +76,41 @@ static string Resolve (this AstType type, EvaluationContext ctx, List<object> ar
...
+		static string Resolve (this RefTypeSyntax type, EvaluationContext ctx, List<object> args)
+		{
+			return type.Type.Resolve (ctx, args);
+		}
 
-			return name;
+		static string Resolve (this NullableTypeSyntax type, EvaluationContext ctx, List<object> args)
+		{
+			args.Insert (0, type.ElementType.Resolve (ctx));
+			return "System.Nullable`1";
```
Therzok, line 92: args unused?

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluator.cs
```diff
@@ -100,15 +102,16 @@ string Resolve (DebuggerSession session, SourceLocation location, string express
 
 			expression = ReplaceExceptionTag (expression, session.Options.EvaluationOptions.CurrentExceptionTag);
 
-			var expr = new CSharpParser ().ParseExpression (expression);
+			
+			var expr = SyntaxFactory.ParseExpression (expression);
```
Therzok, line 106: Btw, this is what VS Windows uses: https://github.com/dotnet/interactive-window

Therzok: Oh, that's not the immediate window, nvm, it's a different repl, I think?

Therzok: Although, this is the entry-point for submitting repl input.

Therzok (commented):
Mostly LGTM. Do things like `Span<int> x = stackalloc int[10]` work?

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs
nosami: resolving a type works now the same way as before, although I have to say I don't really see the point in it.

What's the point of resolving type "System.Console" or namespace "System"

Resolving instances of a type on the other hand makes perfect sense ........

nosami: also, `typeof(System.Console)` is perfectly fine too.

Just not sure what the point of `System.Console` is

Therzok: @nosami `System.Console` should give me static members in that class.

nosami: OK. I guess that does make sense.

I still don't think we should have namespaces though. In fact, the old code used to guess if something was a namespace https://github.com/mono/debugger-libs/blob/f7085b2d81cdf1acb85189428c149ac5b98b6bbc/Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs#L1158

Therzok: Yeah, namespaces make no sense. Static member sof a type do, though.

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs
```diff
@@ -68,6 +68,10 @@ static string ResolveTypeName (SyntaxNode type)
 			string name = type.ToString ();
 			if (name.StartsWith ("global::", StringComparison.Ordinal))
 				name = name.Substring ("global::".Length);
+
+			var indexOfSlash = name.IndexOf('<');
```
Therzok, line 72: slash? :D

nosami: heh... I remember when I wrote this writing a `/` and replacing it with `<` :)

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
```diff
@@ -483,10 +483,10 @@ static string ResolveFullPath (string path)
 			if (string.IsNullOrEmpty(path))
 				return path;
 
-			string result = null;
+			string result;
 			try
 			{
-				result = realpath(path, IntPtr.Zero);
+				result = realpath(path, IntPtr.Zero) ?? Path.GetFullPath(path);
```
Therzok, line 489: Nit: Could just do the Path.GetFullPath on the return statement. Less work for the JIT, and more likely to inline everything.

Therzok:
```csharp
string result = null;
try { result = realpath(path, IntPtr.Zero); catch {}

return string.IsNullOrEmpty(result) ? Path.GetFullPath(path) : result;
```

Therzok (approved):
LGTM!

nosami:
979 Asserts in https://github.com/mono/debugger-libs/blob/main/UnitTests/Mono.Debugging.Tests/Shared/EvaluationTests.cs are now passing

sandyarmstrong (changes requested):
Blocking on using `$(NuGetVersionRoslyn)` and `$(NuGetVersionSystemCollectionsImmutable)` (which vsmac can override) instead of hard-coding package versions. Otherwise love it!

nosami:
> Blocking on using `$(NuGetVersionRoslyn)` and `$(NuGetVersionSystemCollectionsImmutable)` (which vsmac can override) instead of hard-coding package versions. Otherwise love it!

Yeah... I was just going to make sure everything else worked before I changed this :)

Mono.Debugging/Mono.Debugging.csproj
nosami: Done

Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs
nosami: fixed

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs
nosami: renamed to index :)

Mono.Debugging/Mono.Debugging.Evaluation/NRefactoryExpressionEvaluatorVisitor.cs
nosami: This is all working now as before.

mkrueger:
looks good - +1

sandyarmstrong (approved):

PR #344, opened 2021-07-15
[android][debugger] Implement support to debug after hot reload

thaystg:
Related to this PR:
https://github.com/dotnet/runtime/pull/55722

- Invalidate locals_variable, debug_information and locations after apply an update.
- Get new information from runtime.
- Remove the old breakpoint with the wrong offset and create a new one with the correct offset.


lambdageek (approved):

lambdageek:
@joj After this, locals and breakpoints should work after hot reload updates a method

lambdageek:
Since this will disable the breakpoint in the old method body, if the old method was still executing we won't stop in it anymore.  That's probably fine, but we might get surprised customer issues about it since EnC in .net framework works a little differently.  But hot reload is not EnC.

PR #347, opened 2021-08-27
Support debugging even without find types like: DebuggerHiddenAttribute, DebuggerNonUserCodeAttribute, DebuggerTypeProxyAttribute

thaystg:
Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1384996?srWorkItemMention&src-action=artifact_link


thaystg:
@joj do you need to bump it somewhere right?

PR #350, opened 2021-12-18
Proper support for Catchpoint location signature

KirillOsenkov:
When an exception happens we didn't correctly retrieve and store the location of the exception.

We used to append the module name even if the method name already contained the module name (so we'd end up with Module!Module.dll!Class.Method). Also the Address is 0 most of the time for VSCode Debug Protocol, so no need to append it if it's zero.

Finally, store the location signature on the Catchpoint itself, so that when the user selects "Ignore exceptions at this location" we use the exact location that the debugger gave us and add it to the ignore list. Otherwise the exception can be on a different frame (the current frame that has the source code, and not the bottom frame of the stack).

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/Catchpoint.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/Catchpoint.cs
@@ -85,6 +85,17 @@ namespace Mono.Debugging.Client
 			set { includeSubclasses = value; }
 		}
 
+		/// <summary>
+		/// When an exception first happens, we are given the frame of the exception.
+		/// However by the time the user chooses "Ignore this location" in the exception dialog,
+		/// the current frame might be a different frame (the one with source code),
+		/// and not necessarily the bottom frame of the stack.
+		/// Make sure we remember the original frame where the exception happened
+		/// so that when it happens next time, we use the same frame signature to compare
+		/// against.
+		/// </summary>
+		public string CurrentLocationSignature { get; set; }
+
 		public HashSet<string> Ignored { get; private set; } = new HashSet<string> (StringComparer.Ordinal);
 
 		public bool ShouldIgnore(string type, string locationSignature)
--- a/Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
@@ -135,7 +135,16 @@ namespace Mono.Debugging.Client
 		/// </summary>
 		public string GetLocationSignature ()
 		{
-			return $"{Path.GetFileNameWithoutExtension (this.FullModuleName)}!{this.SourceLocation.MethodName}:{this.Address}";
+			string methodName = this.SourceLocation.MethodName;
+			if (methodName != null && !methodName.Contains("!")) {
+				methodName = $"{Path.GetFileName (this.FullModuleName)}!{methodName}";
+			}
+
+			if (this.Address != 0) {
+				methodName = $"{methodName}:{Address}";
+			}
+
+			return methodName;
 		}
 
 		public string GetFullStackFrameText ()


sandyarmstrong (approved):

Mono.Debugging/Mono.Debugging.Client/Catchpoint.cs
```diff
@@ -85,6 +85,17 @@ internal override XmlElement ToXml (XmlDocument doc, string baseDir)
...
+		/// <summary>
+		/// When an exception first happens, we are given the frame of the exception.
+		/// However by the time the user chooses "Ignore this location" in the exception dialog,
+		/// the current frame might be a different frame (the one with source code),
+		/// and not necessarily the bottom frame of the stack.
+		/// Make sure we remember the original frame where the exception happened
+		/// so that when it happens next time, we use the same frame signature to compare
+		/// against.
+		/// </summary>
+		public string CurrentLocationSignature { get; set; }
```
Therzok, line 97: I wonder if we could use a smarter mechanism here, instead of string comparison by assigning IDs (mvid, method token, etc) and associating by that.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
```diff
@@ -135,7 +135,16 @@ internal void Attach (DebuggerSession debugSession)
 		/// </summary>
 		public string GetLocationSignature ()
 		{
-			return $"{Path.GetFileNameWithoutExtension (this.FullModuleName)}!{this.SourceLocation.MethodName}:{this.Address}";
+			string methodName = this.SourceLocation.MethodName;
+			if (methodName != null && !methodName.Contains("!")) {
```
Therzok, line 139: This optimizes to a SIMD contains:

```suggestion
			if (methodName != null && !methodName.Contains('!')) {
```

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
```diff
@@ -135,7 +135,16 @@ internal void Attach (DebuggerSession debugSession)
...
+				methodName = $"{Path.GetFileName (this.FullModuleName)}!{methodName}";
+			}
+
+			if (this.Address != 0) {
+				methodName = $"{methodName}:{Address}";
```
Therzok, line 144: Should the CurrentLocationSignature be a struct { location, method token }?

Therzok (approved):

Mono.Debugging/Mono.Debugging.Client/Catchpoint.cs
Therzok: Yeah, it could just be a struct here, `struct IdentitySignature { FileName, MethodName, Address }`

Therzok (changes requested):
Sorry, marking as requesting changes the more I look at this, having a struct identity signature is probably a cleaner approach.

Mono.Debugging/Mono.Debugging.Client/Catchpoint.cs
KirillOsenkov: No, because different debuggers get different information. VSCode just sends us a string, with module name already prepended. They also don’t give us the IL offset at all. This information needs to be free form, to enable all debuggers, and leave the string format up to each debugger to interpret at will. It’s also future proof.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
KirillOsenkov: No, let’s keep it simple and flexible.

KirillOsenkov:
I’m very strongly for the string approach. I’m confident that a struct is not the right approach here. We don’t need the structure, all we need is an immutable token value that can be easily stored and presented to the user. Moreover, we specifically don’t want to impose any structure on the user. Different debuggers may have different location descriptions. And we’ve lost the battle anyway since the VSCode Protocol just sends us a single unstructured string anyway.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
Therzok: Define flexible here. Because the problem is:

1. You can't define case sensitivity for a language - i.e. VB
2. You can't define case sensitivity based on filesystem

> APFS: Uses the APFS format. Choose this option if you don’t need an encrypted or case-sensitive format.
> APFS (Encrypted): Uses the APFS format and encrypts the volume.
> APFS (Case-sensitive): Uses the APFS format and is case-sensitive to file and folder names. For example, folders named “Homework” and “HOMEWORK” are two different folders.
> APFS (Case-sensitive, Encrypted): Uses the APFS format, is case-sensitive to file and folder names, and encrypts the volume. For example, folders named “Homework” and “HOMEWORK” are two different folders.

Mono.Debugging/Mono.Debugging.Client/StackFrame.cs
KirillOsenkov: This is probably as far from a perf bottleneck as it gets, would really be a microoptimization.

Also .NET Framework (and Mono) don’t have a string.Contains(char), and I’d prefer to not call via an interface.

Therzok (approved):

nosami (approved):

PR #351, opened 2021-12-20
Add net472 back to Mono.Debugging.Soft

pjcollins:
Xamarin.Android has build tasks and tests that target net472 which won't
build without multitargeting Mono.Debugging.Soft.

### DIFF
--- a/Mono.Debugging.Soft/Mono.Debugging.Soft.csproj
+++ b/Mono.Debugging.Soft/Mono.Debugging.Soft.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>..\Mono.Debugging\mono.debugging.snk</AssemblyOriginatorKeyFile>
     <NoWarn>1591;1573</NoWarn>


pjcollins:
@nosami I don't have permission to request reviewers, but hopefully you can take a look at this when you have a moment :)

Therzok (approved):

Mono.Debugging.Soft/Mono.Debugging.Soft.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
```
Therzok, line 4: We're working towards slowly phasing out net472, because I think that implies a dependency on Mono, so I wonder if this should conditionally target net472 and net6.0.

pjcollins: Do you want me to add some property that could control multitargeting this in this current PR, or is this something we should do in a future PR?

Both [Mono.Debugger.Soft](https://github.com/mono/debugger-libs/blob/eaf83aacd4f5b41e240ceb3a61dd49f64227370f/Mono.Debugger.Soft/Mono.Debugger.Soft.csproj#L4) and [Mono.Debugging](https://github.com/mono/debugger-libs/blob/eaf83aacd4f5b41e240ceb3a61dd49f64227370f/Mono.Debugging/Mono.Debugging.csproj#L4) are targeting `net6.0` and `net472`, so I figured this removal of `net472` was maybe just an oversight.

Therzok: I'll deal with it if it becomes a problem then! I saw that `net472` shells out to `mono` when building, so when we remove `mono`, if it breaks, I'll do that.

PR #352, opened 2022-02-07
Fix race condition in conditional debugger tests

mrward:
- Conditions on the breakpoints were set after the breakpoint is added causing the conditions to be ignored depending on how fast the debugger session updates the changes.
 - Update debugger tests to use .NET 6 instead of .NET Core 3.0.

### DIFF
--- a/UnitTests/Mono.Debugging.Tests/Shared/DebugTests.cs
+++ b/UnitTests/Mono.Debugging.Tests/Shared/DebugTests.cs
@@ -257,11 +257,17 @@ namespace Mono.Debugging.Tests
 
 		public Breakpoint AddBreakpoint (string breakpointMarker, int offset = 0, string statement = null, ITextFile file = null)
 		{
+			var bp = CreateBreakpoint (breakpointMarker, offset, statement, file);
+			Session.Breakpoints.Add (bp);
+			return bp;
+		}
+
+		public Breakpoint CreateBreakpoint (string breakpointMarker, int offset = 0, string statement = null, ITextFile file = null)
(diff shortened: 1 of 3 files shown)


Therzok (approved):

UnitTests/Mono.Debugging.Tests/Shared/DebugTests.cs
```diff
@@ -256,12 +256,18 @@ void GetLineAndColumn (string breakpointMarker, int offset, string statement, ou
...
+			return bp;
+		}
+
+		public Breakpoint CreateBreakpoint (string breakpointMarker, int offset = 0, string statement = null, ITextFile file = null)
 		{
 			file = file ?? SourceFile;
 			int col, line;
 			GetLineAndColumn (breakpointMarker, offset, statement, out line, out col, file);
 			var bp = new Breakpoint (file.Name, line, col);
-			Session.Breakpoints.Add (bp);
```
Therzok, line 264: This is so weird, isn't this all synchronous?

Therzok: Ohhh, I get it now.

mrward: The Breakpoint class has a CommitChanges method which would probably be another way to do this, the breakpoints dialog does this when updating an existing breakpoint, but that would likely trigger two updates in the vscode debugger session for the same breakpoint - seemed easier to have everything ready when the breakpoint is added.

PR #354, opened 2022-02-14
Fix Failing Test, Add Batch Methods for Events

marshall-lerner:
The FileRename test in BreakpointsAndSteppingTests uses a relative path to assert equality, whereas with the update to make all paths absolute, we need for the path to be absolute.

Also, Whenever there is a call to any of [these events](https://github.com/mono/debugger-libs/blob/main/Mono.Debugging/Mono.Debugging.Client/BreakpointStore.cs#L594-L609), there is unnecessary work on the receiving side when there are multiple breakpoints that change.

See https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1476114

### DIFF
--- a/UnitTests/Mono.Debugging.Tests/Shared/BreakpointsAndSteppingTests.cs
+++ b/UnitTests/Mono.Debugging.Tests/Shared/BreakpointsAndSteppingTests.cs
@@ -1315,7 +1315,7 @@ namespace Mono.Debugging.Tests
 			store.FileRenamed ("fileName1.cs", "fileName2.cs");
 			var bps = store.GetBreakpoints ();
 			Assert.AreEqual (1, bps.Count);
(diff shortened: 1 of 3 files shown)


UnitTests/Mono.Debugging.Tests/Shared/BreakpointsAndSteppingTests.cs
```diff
@@ -1315,7 +1315,7 @@ public void FileRename()
 			store.FileRenamed ("fileName1.cs", "fileName2.cs");
 			var bps = store.GetBreakpoints ();
 			Assert.AreEqual (1, bps.Count);
-			Assert.AreEqual ("fileName2.cs", bps [0].FileName);
+			Assert.AreEqual (Environment.CurrentDirectory + "/fileName2.cs", bps [0].FileName);
```
mrward, line 1318: Could do a Path.Combine here instead of using the forward slash.

mrward (approved):

PR #355, opened 2022-02-17
Make sure that the only real paths get resolved

marshall-lerner:
We were encountering an exception with Functional Breakpoints, which have no filepath, because ResolveFullPath was throwing


marshall-lerner:
@mrward do you think it's better to put some conditions in the ResolveFullPath, or catch when a filepath is null or empty before we call it?

mrward:
I would probably move the string.IsNullOrEmpty check to the ResolveFullPath method. Change is one place and then anybody can call ResolveFullPath without having to worry about checking.

Therzok (approved):

PR #360, opened 2022-04-12
Attempt to fix deadlock

nosami:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1493044/

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -1,4 +1,4 @@
-// DebuggerSession.cs
+﻿// DebuggerSession.cs
 //
 // Author:
 //   Ankit Jain <jankit@novell.com>
@@ -32,6 +32,7 @@ using System.Collections.Generic;
 
 using Mono.Debugging.Backend;
 using Mono.Debugging.Evaluation;
+using System.Collections.Concurrent;
 
 namespace Mono.Debugging.Client
 {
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -352,42 +353,32 @@ public void SetOutputOptions(OutputOptions options)
...
-								}
-								lock (slock) {
-									try {
-										actionToExecute ();
-									} catch (Exception ex) {
-										HandleException (ex);
-									}
+				actionsQueue.Enqueue (action);
+				if (!threadExecuting) {
+					threadExecuting = true;
```
Therzok, line 364: This is now race condition prone.

Therzok: There are at least 3 operations at hand: Load + if + store

nosami: yeah... I didn't think about this boolean set. Need to see why that is needed

Therzok: Interlocked.CompareExchange might be useful here, but there are 2 objects that need to be synchronized - the boolean and the ConcurrentQueue.

That means we need to either:
1. Loop over setting the boolean with InterlockedCompareExchange - as it can "lose" the set because something else set it first - and then trigger the enqueue.
2. Use a `ConcurrentExclusiveSchedulerPair` and queue up on the tasks on the ExclusiveScheduler - but from the looks of it it is still synchronously locking
3. Use BlockingCollection instead? But that might also synchronously block.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -352,42 +353,32 @@ public void SetOutputOptions(OutputOptions options)
...
+					ThreadPool.QueueUserWorkItem (delegate {
+						while (true) {
+							if (actionsQueue.TryDequeue (out var actionToExecute)) {
```
Therzok, line 367: Nit: `while (actionsQueue.TryDequeue(...))`

nosami:
Closing this as I can't verify if it actually fixes anything and I potentially introduced bugs here.

PR #361, opened 2022-05-02
Protect from errors related to cancelling debugger connection

mauroa:
If any unexpected error occurs when ending the debugger launch session, some important things may not happen like invoking the TargetExited event, which could cause the debugger session to hang.
For this reason, we should handle any exception happening around it and report it accordingly
One reason for this problem to reproduce is when the debugger session launches but the connection with the app is not established, so the session will never end and potentially hang under any UI interaction

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs
@@ -338,6 +338,8 @@ namespace Mono.Debugger.Soft
 
 		public static void CancelConnection (IAsyncResult asyncResult)
 		{
+			//AsyncState could be null if the debugger incoming connection doesn't happen, so there's no socket between debugger and debuggee
+			//This could occur if the debugger session started but the connection to the app failed at some point
 			((Socket)asyncResult.AsyncState)?.Close ();
 		}
 		
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs
@@ -338,7 +338,7 @@ namespace Mono.Debugger.Soft
 
 		public static void CancelConnection (IAsyncResult asyncResult)
 		{
-			((Socket)asyncResult.AsyncState).Close ();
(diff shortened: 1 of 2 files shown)


Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs
```diff
@@ -338,7 +338,7 @@ public static async Task<VirtualMachine> ConnectInternalAsync (Socket dbg_sock,
 
 		public static void CancelConnection (IAsyncResult asyncResult)
 		{
-			((Socket)asyncResult.AsyncState).Close ();
+			((Socket)asyncResult.AsyncState)?.Close ();
```
Therzok, line 341: What sets the asyncstate to null?

mauroa: It happens at some point in the same `VirtualMachineManager` class because the method that returns the `IAsyncResult` (BeginListen) doesn't end up correctly.

Here it starts listening for connections: https://github.com/mono/debugger-libs/blob/main/Mono.Debugging.Soft/SoftDebuggerSession.cs#L247

Here it calls an internal listen method: https://github.com/mono/debugger-libs/blob/cdf25369c56e6006f64b63fd3ea4d817c2cf8da8/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs#L240

Here it ends up abnormally: https://github.com/mono/debugger-libs/blob/cdf25369c56e6006f64b63fd3ea4d817c2cf8da8/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachineManager.cs#L183

Then, we are catching an exception on the XamarinVS side and calling the `EndSession` method of the debugger session, which at that point has the `IAsyncResult.AsyncState` as null (presumably because it never completed correctly the `BeginListen` method).

Therzok: Thanks for the clarification. Can please we add that info in code comments comments as well? :D

mauroa: Done

thaystg (approved):

PR #363, opened 2022-06-01
initial Module View PR

jonathan34c:
Create Assembly data class and add event listener to Debugger Sesson


sgmunn (approved):

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -49,6 +49,7 @@ public abstract class DebuggerSession: IDisposable
 	{
 		readonly Dictionary<BreakEvent, BreakEventInfo> breakpoints = new Dictionary<BreakEvent, BreakEventInfo> ();
 		readonly Dictionary<string, string> resolvedExpressionCache = new Dictionary<string, string> ();
+		public List<Assembly> assemblies = new List<Assembly> ();
```
slluis, line 52: This is not right. It should be a read-only property, not a public field.

Therzok: @slluis it was merged into jonathan's other branch.

PR #364, opened 2022-06-09
Add symbol server options and tweak cache path format

nosami:
Part of work to make sourcelink work for dotnet apps and add symbol server support

https://github.com/xamarin/vsmac/pull/7167

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
@@ -1,17 +1,11 @@
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
-using System.Text;
 
 namespace Mono.Debugging.Client
 {
 	[Serializable]
 	public class SourceLink
 	{
-		// Keep the / character to use as a path separator to help group related symbols
-		static readonly HashSet<char> invalids = new HashSet<char>(Path.GetInvalidFileNameChars ().Except (new char[] { '/' }));
-
 		public string Uri { get; }
 
 		public string RelativeFilePath { get; }
@@ -25,24 +19,7 @@ namespace Mono.Debugging.Client
 		public string GetDownloadLocation (string cachePath)
 		{
 			var uri = new Uri (Uri);
-			return Path.Combine (cachePath, MakeValidFileName (uri));
-		}
-
-		static string MakeValidFileName (Uri uri)
-		{
-			// Remove scheme from uri
-			var text = uri.Host + uri.PathAndQuery;
-			var sb = new StringBuilder (text.Length);
-			for (int i = 0; i < text.Length; i++) {
-				char c = text[i];
-				if (invalids.Contains (c)) {
-					sb.Append ('_');
-				} else
-					sb.Append (c);
-			}
-			if (sb.Length == 0)
-				return "_";
-			return sb.ToString ();
+			return Path.Combine (cachePath, uri.Host + uri.PathAndQuery);
 		}
 	}
 }
--- a/Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
@@ -1,12 +1,19 @@
-﻿using System;
+using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Text;
 
 namespace Mono.Debugging.Client
 {
 	[Serializable]
 	public class SourceLink
 	{
+		// Keep the / character to use as a path separator to help group related symbols
+		static readonly HashSet<char> invalids = new HashSet<char>(Path.GetInvalidFileNameChars ().Except (new char[] { '/' }));
+
 		public string Uri { get; }
+
 		public string RelativeFilePath { get; }
 
 		public SourceLink (string uri, string relativeFilePath)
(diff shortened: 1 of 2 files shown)


Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
```diff
@@ -1,12 +1,19 @@
...
+using System.Linq;
+using System.Text;
 
 namespace Mono.Debugging.Client
 {
 	[Serializable]
 	public class SourceLink
 	{
+		// Keep the / character to use as a path separator to help group related symbols
+		static readonly HashSet<char> invalids = new HashSet<char>(Path.GetInvalidFileNameChars ().Except (new char[] { '/' }));
```
Therzok, line 13: I think an array here is probably going to be faster than a hashset.

Therzok: But it's not a big deal.

Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
```diff
@@ -17,7 +24,25 @@ public SourceLink (string uri, string relativeFilePath)
...
 		{
-			return Path.Combine (cachePath, new Uri (Uri).GetComponents (UriComponents.Path, UriFormat.SafeUnescaped));
+			var uri = new Uri (Uri);
+			return Path.Combine (cachePath, MakeValidFileName (uri));
+		}
+
+		static string MakeValidFileName (Uri uri)
+		{
+			// Remove scheme from uri
+			var text = uri.Host + uri.PathAndQuery;
```
Therzok, line 34: We're using a string builder to avoid allocations, but we pay up front for the allocation of this string?

Therzok (approved):

Mono.Debugging/Mono.Debugging.Client/SourceLink.cs
nosami: it's a lookup table that is used repeatedly, so hashset is a better choice here

Therzok: Have you _profiled_ it? I assume not, so here are some results that [say the opposite](https://github.com/Therzok/BenchPlayground/blob/8377c906a5a4aa2f1f7472b67a0e15c3dfcf713e/BenchPlayground/Benchmarks/System.Collections/ArrayVsHashSetStrings.md).

You have to compute a hashcode to get a bucket then do equality comparison for a HashSet. For simple arrays like `char[]`, `Array.IndexOf` does a SIMD scan.

nosami: I actually assumed that System.IO.Path.GetInvalidFileNameChars () would contain many more than 2 items :)

nosami: but also, your profile results are only valid for the exact number of items in the array that you use (5)

nosami: > I actually assumed that System.IO.Path.GetInvalidFileNameChars () would contain many more than 2 items :)
> 
> ```
> csharp> System.IO.Path.GetInvalidFileNameChars ()    
> { '\x0', '/' } 
> ```
> 
> If I'd realised that there were only 2 items, I would have never used a HashSet. No need for profile results here :D

This actually makes me question whether or not MakeValidFileName(uri) could just be => uri.Host + uri.PathAndQuery; as I don't think \x0 is a possible character in uri.Host or uri.PathAndQuery

Therzok: debugger-libs should still run on Windows, which returns a bit different values:
https://sharplab.io/#v2:C4LgTgrgdgPgAgJgAwFgBQiCM65IARyYB0AkgPIDc66AZgPZgCmAhgMYAWeAFAG7Nh4OeAJZQ8ABWbB2RAOKNgJKHwA2wgCYAxYSsYA5ZgFtGAYXb8AzlwCU1vAG90eZ4QCcXAET2kAXweYfDwAaQXYQri4AIwBPYEZrDlsKPHQfdCA=

Therzok: > but also, your profile results are only valid for the exact number of items in the array that you use (5)

It's valid for a lot of values until 64, from what I recall, there are other benchmarks in that repo that are relevant.

nosami: Indeed, MakeValidFileName is not even needed at all

nosami: > debugger-libs should still run on Windows, which returns a bit different values:
> https://sharplab.io/#v2:C4LgTgrgdgPgAgJgAwFgBQiCM65IARyYB0AkgPIDc66AZgPZgCmAhgMYAWeAFAG7Nh4OeAJZQ8ABWbB2RAOKNgJKHwA2wgCYAxYSsYA5ZgFtGAYXb8AzlwCU1vAG90eZ4QCcXAET2kAXweYfDwAaQXYQri4AIwBPYEZrDlsKPHQfdCA=

grr.. I forgot it needs to work on Windows. Looks like 41 results. I would be very surprised if Array came out on top here

nosami: Even on Windows, MakeValidFileName is probably not needed. I don't think there are any URI characters that are invalid file characters.

PR #366, opened 2022-06-22
Provide listed assembly in process for moduleview

jonathan34c:
1. Add assembly data structure,
2. Store assemblies that were loaded in the debugger session.
3. Allow the process to access loaded assemblies in the debugger session.
4. Allow Debbuggng service to access the loaded assemblies to create modules pad

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -2262,19 +2262,17 @@ namespace Mono.Debugging.Soft
 
 		private void HandleAssemblyLoaded (AssemblyMirror asm)
 		{
-			var symbolStatus = string.Empty;
-			var assemblyName = string.Empty;
-			var hasSymbol = false;
 			var name = asm.GetName ();
 			var assemblyObject = asm.GetAssemblyObject ();
-			if (!asm.IsDynamic) {
+			bool isDynamic = asm.IsDynamic;
+			string assemblyName;
+			bool hasSymbol;
+			if (!isDynamic) {
 				var metaData = asm.GetMetadata ();
-				symbolStatus = metaData.MainModule.HasSymbols == true ? "Symbol loaded" : "Skipped loading symbols";
 				assemblyName = metaData.MainModule.Name;
 				hasSymbol = metaData.MainModule.HasSymbols;
 			} else {
-				symbolStatus = "Skipped loading symbol (dynamic)";
-				assemblyName = "Dynamic assembly";
+				assemblyName = string.Empty;
 				hasSymbol = false;
 			}
 			var assembly = new Assembly (
@@ -2282,16 +2280,18 @@ namespace Mono.Debugging.Soft
 					asm.Location,
 					true,
 					hasSymbol,
-					symbolStatus,
-					"",
+					string.Empty,
+					string.Empty,
 					-1,
 					name.Version.Major.ToString (),
 					// TODO: module time stamp
-					"",
+					string.Empty,
 					assemblyObject.Address.ToString (),
 					string.Format ("[{0}]{1}", asm.VirtualMachine.TargetProcess.Id, asm.VirtualMachine.TargetProcess.ProcessName),
 					asm.Domain.FriendlyName,
-					asm.VirtualMachine.TargetProcess.Id
+					asm.VirtualMachine.TargetProcess.Id,
+					hasSymbol,
+					isDynamic
 			);
 
 			OnAssemblyLoaded (assembly);
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -49,6 +49,7 @@ using Mono.Debugging.Evaluation;
 
 using StackFrame = Mono.Debugger.Soft.StackFrame;
 using System.Collections.Immutable;
+using Assembly = Mono.Debugging.Client.Assembly;
 
 namespace Mono.Debugging.Soft
 {
@@ -2248,9 +2249,8 @@ namespace Mono.Debugging.Soft
 			if (events.Length > 1 && events.Any (a => a.Assembly != asm))
 				throw new InvalidOperationException ("Simultaneous AssemblyLoadEvent for multiple assemblies");
 
-			OnAssemblyLoaded(asm.Location);
-
-			RegisterAssembly(asm);
+			HandleAssemblyLoaded (asm);
+			RegisterAssembly (asm);
 			bool isExternal;
 			isExternal = !UpdateAssemblyFilters (asm) && userAssemblyNames != null;
 
@@ -2258,6 +2258,43 @@ namespace Mono.Debugging.Soft
 				string flagExt = isExternal ? " [External]" : "";
 				OnDebuggerOutput (false, string.Format ("Loaded assembly: {0}{1}\n", asm.Location, flagExt));
 			}
+		}
+
+		private void HandleAssemblyLoaded (AssemblyMirror asm)
+		{
+			var symbolStatus = string.Empty;
+			var assemblyName = string.Empty;
+			var hasSymbol = false;
+			var name = asm.GetName ();
+			var assemblyObject = asm.GetAssemblyObject ();
+			if (!asm.IsDynamic) {
+				var metaData = asm.GetMetadata ();
+				symbolStatus = metaData.MainModule.HasSymbols == true ? "Symbol loaded" : "Skipped loading symbols";
+				assemblyName = metaData.MainModule.Name;
+				hasSymbol = metaData.MainModule.HasSymbols;
+			} else {
+				symbolStatus = "Skipped loading symbol (dynamic)";
+				assemblyName = "Dynamic assembly";
+				hasSymbol = false;
+			}
+			var assembly = new Assembly (
+					assemblyName,
+					asm.Location,
+					true,
+					hasSymbol,
+					symbolStatus,
+					"",
+					-1,
+					name.Version.Major.ToString (),
+					// TODO: module time stamp
+					"",
+					assemblyObject.Address.ToString (),
+					string.Format ("[{0}]{1}", asm.VirtualMachine.TargetProcess.Id, asm.VirtualMachine.TargetProcess.ProcessName),
+					asm.Domain.FriendlyName,
+					asm.VirtualMachine.TargetProcess.Id
+			);
+
+			OnAssemblyLoaded (assembly);
 		}
 
 		void RegisterAssembly (AssemblyMirror asm)
--- a/Mono.Debugging/Mono.Debugging.Client/Assembly.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/Assembly.cs
@@ -44,7 +44,7 @@ namespace Mono.Debugging.Client
 			UserCode = userCode;
 			ProcessId = processId;
 			IsDynamic = isDynamic;
-			HasSymbol = hasSymbol;
+			HasSymbols = hasSymbol;
 		}
 		public Assembly (string path)
 		{
@@ -77,7 +77,7 @@ namespace Mono.Debugging.Client
 
 		public long? ProcessId { get; private set; } = -1;
 
-		public bool HasSymbol { get; private set; }
+		public bool HasSymbols { get; private set; }
 
 		public bool IsDynamic { get; private set; }
 	}
--- a/Mono.Debugging/Mono.Debugging.Client/Assembly.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/Assembly.cs
@@ -1,0 +1,79 @@
+//
+// Assembly.cs
+//
+// Author:
+//       Jonathan Chang <t-jochang@microsoft.com>
+//
+// Copyright (c) 2022 
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+namespace Mono.Debugging.Client
+{
+	public class Assembly
+	{
+		public Assembly (string name, string path, bool optimized, bool userCode, string symbolStatus, string symbolFile, int? order, string version, string timestamp, string address, string process, string appdomain, long? processId)
+		{
+			Name = name;
+			Path = path;
+			Optimized = optimized;
+			SymbolStatus = symbolStatus;
+			SymbolFile = symbolFile;
+			Order = order.GetValueOrDefault (-1);
+			TimeStamp = timestamp;
+			Address = address;
+			Process = process;
+			AppDomain = appdomain;
+			Version = version;
+			UserCode = userCode;
+			ProcessId = processId;
+
+		}
+		public Assembly (string path)
+		{
+			Path = path;
+		}
+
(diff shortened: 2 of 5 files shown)


Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2245,7 +2246,38 @@ void HandleAssemblyLoadEvents (AssemblyLoadEvent[] events)
 			if (events.Length > 1 && events.Any (a => a.Assembly != asm))
 				throw new InvalidOperationException ("Simultaneous AssemblyLoadEvent for multiple assemblies");
 
-			OnAssemblyLoaded(asm.Location);
+			// todo: fil these in
```
sgmunn, line 2249: if these are done, we should remove the todo's

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2245,7 +2246,38 @@ void HandleAssemblyLoadEvents (AssemblyLoadEvent[] events)
...
+				asm.Location,
+				true,
+				asm.GetMetadata ().MainModule.HasSymbols,
+				symbolStatus,
+				"",
+				-1,
+				asm.GetName ().Version.Major.ToString(),
+				asm.VirtualMachine.Process.StartTime.ToUniversalTime().ToString(),
+				asm.GetAssemblyObject ().Address.ToString(),
+				"["+asm.VirtualMachine.TargetProcess.Id +"]"+asm.VirtualMachine.TargetProcess.ProcessName,
```
sgmunn, line 2276: Could probably use a string.Format() here

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -300,6 +301,14 @@ public void SetOutputOptions(OutputOptions options)
 			get; private set;
 		}
 
+		public List<Assembly> Assemblies {
+			get {
+				lock (assemblies) {
+					return assemblies;
+				}
```
sgmunn, line 308: returning the underlying list itself would allow the session to mutate the list in the background which would not be ideal for consumers. instead we can return an array and then simply return assemblies.ToArray() inside the lock

slluis:
Please set a proper title and description of the PR

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -1300,11 +1309,15 @@ internal protected void OnTargetDebug (int level, string category, string messag
...
 
-		internal protected void OnAssemblyLoaded (string assemblyLocation)
+		internal protected void OnAssemblyLoaded (Assembly assembly)
 		{
-			AssemblyLoaded?.Invoke (this, new AssemblyEventArgs (assemblyLocation));
+			lock (assemblies) {
+				assemblies.Add (assembly);
+			}
+
+			AssemblyLoaded?.Invoke (this, new AssemblyEventArgs (assembly.Path));
```
slluis, line 1318: It would be better to add a reference to the Assembly instance to AssemblyEventArgs, instead of just the assembly path.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
sgmunn: @slluis   Would the ProcessInfo, in I presume currentProcesses, always be set? From what I can tell this is only set if the client calls GetProcesses()?

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
slluis: The existing event doesn't need to change, a new property and constructor can be added to AssemblyEventArgs

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
slluis: Yes, currentProcesses is filled when GetProcesses() is called, but that doesn't matter. We can keep the list of assemblies internally in DebuggerSession, and in ProcessInfo just get the subset corresponding to the process, something similar to what it is being done for GetThreads().

Therzok: This should be a method, not a property, as per @slluis' initial request

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,122 @@
...
+			Process = process;
+			AppDomain = appdomain;
+			Version = version;
+			usercode = userCode;
+		}
+		public Assembly (string path)
+		{
+			Path = path;
+		}
+		public string Name {
```
Therzok, line 68: Minor nit. Then all of these don't really need to have backing fields, nor implementations, since the compiler will auto-generate the same code.

```suggestion
		public string Name { get; protected set; }		
```

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -307,6 +307,13 @@ public void SetOutputOptions(OutputOptions options)
 			}
 		}
 
+		public Assembly[] GetAssemblies(long processId)
+		{
+			lock (assemblies) {
+				return assemblies.ToArray ();
```
sgmunn, line 313: this method needs to filter based on the process id that is passed in

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -300,6 +302,19 @@ public void SetOutputOptions(OutputOptions options)
 			get; private set;
 		}
 
+		public Assembly[] GetAssemblies() {
+			lock (assemblies) {
+				return assemblies.ToArray ();
+			}
+		}
+
+		public Assembly[] GetAssemblies(long processId)
```
slluis, line 311: This method should be internal, since getting the assemblies for a process is done through the ProcessInfo object.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -1300,11 +1315,15 @@ internal protected void OnTargetDebug (int level, string category, string messag
 			}
 		}
 
-		internal protected void OnAssemblyLoaded (string assemblyLocation)
+		internal protected void OnAssemblyLoaded (Assembly assembly)
```
slluis, line 1318: This is an API break. Instead of changing the parameter, add a new overload that takes the Assembly as argument. The old overload can call the new one.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -300,6 +302,19 @@ public void SetOutputOptions(OutputOptions options)
 			get; private set;
 		}
 
+		public Assembly[] GetAssemblies() {
```
slluis, line 305: Add documentation.

slluis (changes requested):
Please add a proper title and a proper description to the pull request.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -389,16 +387,16 @@ void EndLaunch ()
...
+				try {
+					if (startArgs != null && startArgs.ConnectionProvider != null) {
+						startArgs.ConnectionProvider.CancelConnect(connection);
+						startArgs = null;
+					} else {
+						VirtualMachineManager.CancelConnection(connection);
+					}
+				} catch(Exception e) {
+					DebuggerLoggingService.LogError("Unhandled error canceling the debugger connection", e);
+				}
```
mrward, line 399: This code looks the same. Did line endings change?

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -641,16 +639,16 @@ void RegisterUserAssemblies (SoftDebuggerStartInfo dsi)
...
-
+		public void AddOrReplaceAssemblyPathMap(string assembly, string path)
+		{
+			assemblyPathMap[assembly] = path;
+		}
+
+		public void AddOrReplaceSymbolPathMap (string assembly, string path)
+		{
+			symbolPathMap[assembly] = path;
+		}
```
mrward, line 650: Similar here. Code is the same but showing up as modified?

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -1302,9 +1323,23 @@ internal protected void OnTargetDebug (int level, string category, string messag
 
 		internal protected void OnAssemblyLoaded (string assemblyLocation)
 		{
-			AssemblyLoaded?.Invoke (this, new AssemblyEventArgs (assemblyLocation));
+			var assembly = new Assembly (assemblyLocation);
+			lock (assemblies) {
```
slluis, line 1327: To avoid code duplication, you can replace these 5 lines by `OnAssemblyLoaded(assembly)`

Mono.Debugging/Mono.Debugging.Client/AssemblyEventArgs.cs
```diff
@@ -30,13 +30,18 @@ namespace Mono.Debugging.Client
 {
 	public class AssemblyEventArgs : EventArgs
 	{
-		public AssemblyEventArgs (string location)
```
slluis, line 33: This constructor can't be removed, it would be a public api break. You can keep it, and create an Assembly instance here with just the location.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -3382,22 +3380,7 @@ public void ApplyChanges (ModuleMirror module, byte[] metadataDelta, byte[] ilDe
 			else
 				pdbArray = rootDomain.CreateByteArray (pdbDelta);
 
-			module.Assembly.ApplyChanges_DebugInformation (metadataDelta, pdbDelta);
```
slluis, line 3385: It looks like this removal and others above are the result of a bad rebase or merge.

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,84 @@
...
+			Process = process;
+			AppDomain = appdomain;
+			Version = version;
+			UserCode = userCode;
+			ProcessId = processId;
+			
+		}
+		public Assembly (string path)
+		{
+			Path = path;
```
sgmunn, line 55: for consistency with the other constructor, you should set Order to -1 as well IMO

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2226,35 +2224,67 @@ void HandleBreakEventSet (Event[] es, bool dequeuing)
...
+					"",
+					-1,
+					name.Version.Major.ToString (),
+					// TODO: module time stamp
+					"",
+					assemblyObject.Address.ToString (),
+					string.Format ("[{0}]{1}", asm.VirtualMachine.TargetProcess.Id, asm.VirtualMachine.TargetProcess.ProcessName),
+					asm.Domain.FriendlyName,
+					asm.VirtualMachine.TargetProcess.Id
+			);
```
Therzok, line 2279: Can we move this into its own method?

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,84 @@
...
+			Name = name;
+			Path = path;
+			Optimized = optimized;
+			SymbolStatus = symbolStatus;
+			SymbolFile = symbolFile;
+			if (order == null) {
+				Order = -1;
+			} else {
+				Order = (int)order;
+			}
```
mrward, line 42: Minor: This could be replaced by:

    Order = order.GetValueOrDefault(-1);

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -2226,53 +2224,88 @@ void HandleBreakEventSet (Event[] es, bool dequeuing)
...
+			var name = asm.GetName ();
+			var assemblyObject = asm.GetAssemblyObject ();
+			if (!asm.IsDynamic) {
+				var metaData = asm.GetMetadata ();
+				symbolStatus = metaData.MainModule.HasSymbols == true ? "Symbol loaded" : "Skipped loading symbols";
+				assemblyName = metaData.MainModule.Name;
+				hasSymbol = metaData.MainModule.HasSymbols;
+			} else {
+				symbolStatus = "Skipped loading symbol (dynamic)";
+				assemblyName = "Dynamic assembly";
```
mrward, line 2273: General question here, how are we handling translations? I am assuming the above text is being displayed in VS Mac? Are we re-translating these over there?

Just wondering if we should avoid having strings here and instead of an enum for symbol status and then generate the strings on the VS Mac side.

Mono.Debugging.Soft/SoftDebuggerSession.cs
```diff
@@ -134,16 +137,16 @@ protected override void OnRun (DebuggerStartInfo startInfo)
 				throw new ArgumentException ("StartArgs has no ConnectionProvider");
 			}
 		}
-		
+
```
slluis, line 140: All whitespace changes in this file need to be reverted.

Mono.Debugging.Soft/SoftDebuggerSession.cs
sgmunn: We might have to add an additional field to the assembly class to indicate that it's dynamic and maybe something to indicate if symbols were loaded or not. then we can assign these strings in the modules pad.

mrward: I guess two bool properties here HasSymbols and IsDynamic would be enough. Then we do not need any text to be specified here.

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,84 @@
...
+
+		public string Name { get; private set; }
+
+		public string Path { get; private set; }
+
+		public bool Optimized { get; private set; }
+
+		public bool UserCode { get; private set; }
+
+		public string SymbolStatus { get; private set; }
```
mrward, line 62: Do we need SymbolStatus now?

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,84 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+namespace Mono.Debugging.Client
+{
+	public class Assembly
+	{
+		public Assembly (string name, string path, bool optimized, bool userCode, string symbolStatus, string symbolFile, int? order, string version, string timestamp, string address, string process, string appdomain, long? processId, bool hasSymbol = false, bool isDynamic = false)
```
slluis, line 31: Please add documentation to this class and its properties.

slluis (approved):

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,117 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+namespace Mono.Debugging.Client
+{
+	public class Assembly
+	{
+		/// <summary>
+		/// Represent the assembly loaded during the debugging session. This dataclass will be used in the modules pad view for display.
```
sgmunn, line 32: > This dataclass will be used in the modules pad view for display.

this isn't relevant here, since other projects that use this library will have no context of vsmac.

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,117 @@
...
+			ProcessId = processId;
+			IsDynamic = isDynamic;
+			HasSymbols = hasSymbol;
+		}
+		public Assembly (string path)
+		{
+			Path = path;
+		}
+
+		public string Name { get; private set; }
```
sgmunn, line 87: These properties should be documented

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
```diff
@@ -0,0 +1,117 @@
...
+// THE SOFTWARE.
+using System;
+namespace Mono.Debugging.Client
+{
+	public class Assembly
+	{
+		/// <summary>
+		/// Represent the assembly loaded during the debugging session. This dataclass will be used in the modules pad view for display. 
+		/// </summary>
+		/// <param name="name"></param>
```
sgmunn, line 34: You can probably omit the "String" / "Boolean" etc from the documentation here and just keep it like

" Represents the name of the assembly"

Mono.Debugging/Mono.Debugging.Client/Assembly.cs
sgmunn: yes, dotnet core provides this

Therzok (approved):
LGTM

sgmunn (approved):

PR #371, opened 2022-07-20
Remove nested locks on slock

nosami:
There are a few places in DebuggerSession that look like this

```cs
lock(slock) {
    Dispatch(delegate { .... })
}
```

but `Dispatch` is also locking on the same object in a separate thread.

```cs
ThreadPool.QueueUserWorkItem (delegate {
	lock (slock) {
		try {
			actionToExecute ();
		} catch (Exception ex) {
			HandleException (ex);
		}
	}
});
```

When I do something similar in a simplified application, it deadlocks immediately.


nosami:
I moved any code that calls `OnRunning` to be inside the Dispatch methods. This method sets a field to indicate the running state and also fires an event for subscribers.
This means this bit of code is now running on a background thread but I haven't noticed any ill side effects from doing this.

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -409,22 +409,20 @@ public void Run (DebuggerStartInfo startInfo, DebuggerSessionOptions options)
...
-						if (!exceptionHandled)
-							throw;
-					}
-				});
-			}
+
+			Dispatch (delegate {
+				try {
+					this.options = options;
+					OnRunning ();
```
Therzok, line 416: I think this works. If the callbacks in the TargetStarted handler do end up calling other methods which require the lock, they will still have it, because it's the same thread owning the data. I can't think of a way to break this.

Therzok (approved):

mrward (approved):

Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -409,22 +409,20 @@ public void Run (DebuggerStartInfo startInfo, DebuggerSessionOptions options)
...
-						ForceExit ();
-						if (!exceptionHandled)
-							throw;
-					}
-				});
-			}
+
+			Dispatch (delegate {
+				try {
+					this.options = options;
```
slluis, line 415: The options assignment should be done before the Dispatch, since otherwise the Options and EvaluationOptions properties will not return the right value just after calling Run().

Therzok: If we need a thread-safe way to set it, we can use `Interlocked.Exchange(ref this.options, options)`

Therzok: And that way we also avoid the need for a lock for synchronization

PR #372, opened 2022-07-20
Set options before Dispatch

nosami:
This is a feedback fix in response to https://github.com/mono/debugger-libs/pull/371#discussion_r925712902

cc @slluis @Therzok

### DIFF
--- a/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
+++ b/Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
@@ -410,9 +410,10 @@ namespace Mono.Debugging.Client
 			if (options == null)
 				throw new ArgumentNullException (nameof (options));
 
+			Interlocked.Exchange (ref this.options, options);
+
 			Dispatch (delegate {
 				try {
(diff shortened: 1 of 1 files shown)


Mono.Debugging/Mono.Debugging.Client/DebuggerSession.cs
```diff
@@ -410,9 +410,10 @@ public void Run (DebuggerStartInfo startInfo, DebuggerSessionOptions options)
 			if (options == null)
 				throw new ArgumentNullException (nameof (options));
 
+			Interlocked.Exchange (ref this.options, options);
```
Therzok, line 413: Doesn't AttachToProcess need the same change?

Therzok: Yeah, it does, it was also changed in the previous PR

nosami: Thanks, fixed

Therzok (approved):

PR #373, opened 2022-08-01
soft debugger null check for Preview 17.4 version 1

jonathan34c:
Add Null check for the mono project when adding assembly 
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1584183

### DIFF
--- a/Mono.Debugging.Soft/SoftDebuggerSession.cs
+++ b/Mono.Debugging.Soft/SoftDebuggerSession.cs
@@ -2268,28 +2268,28 @@ namespace Mono.Debugging.Soft
 			string assemblyName;
 			bool hasSymbol;
 			if (!isDynamic) {
-				var metaData = asm.GetMetadata ();
-				assemblyName = metaData.MainModule.Name;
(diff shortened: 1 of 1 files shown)


mrward (approved):

Therzok:
The change is not really correct? We use asm before the first line in each condition?

Just do a null check at the top?

jonathan34c:
This method is use under [HandleAssemblyLoadEvent](https://github.com/mono/debugger-libs/blob/91601adb93bb1e884abcec95e48a64f99ff764b4/Mono.Debugging.Soft/SoftDebuggerSession.cs#L1863)
and the [Register assembly method](https://github.com/mono/debugger-libs/blob/91601adb93bb1e884abcec95e48a64f99ff764b4/Mono.Debugging.Soft/SoftDebuggerSession.cs#L2300) also use it as it is. So it's probably safe to assume that the asm will not be null. 
We are more concern if the properties of the asm will be null.

Therzok:
Can we fix that up? There's a correctness issue here, because we use `asm` without the null check, so the null checks are redundant - as the compiler will still emit them.

PR #375, opened 2022-11-21
Trying to avoid getting metadata of all the assemblies loaded from runtime

thaystg:
This should avoid sending all metadata information over the wire which may impact the performance of debugging a maui app.
Also this should avoid crashing while debugging using meadow.


Redth:
I was able to validate that this does fix the Meadow debugger issue. 


I did briefly try testing this in VSMac but I wasn't able to conclusively determine if it helps the debugger session performance. There may be some caching I'm unaware of there. In any case it certainly did not make anything worse and logically it should improve performance and memory consumption since it's no longer sending across all the assembly binaries. 

I was not able to test in VS Windows. 

This looks like a good change!

mauroa (approved):

PR #382, opened 2023-02-23
Move symbol server support from debugger-vs and complete the implementation

thaystg:
I need these informations to correct create the URL to download symbols from symbol server.

I created a new message to avoid passing all the assembly content over the protocol and read it using PEReader which would be easier, but it would spend more time and more memory on client side also.

Related to PRs on runtime and debugger-vs repo:
https://github.com/dotnet/runtime/pull/82555
https://github.com/xamarin/debugger-vs/pull/315

https://user-images.githubusercontent.com/4503299/221028994-a9e5c64b-ae8f-4c85-865e-8d3631d775db.mp4

### DIFF
--- a/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -569,7 +569,8 @@ namespace Mono.Debugger.Soft
 			GET_METHOD_FROM_TOKEN = 12,
 			HAS_DEBUG_INFO = 13,
 			GET_CATTRS = 14,
-			GET_DEBUG_INFORMATION = 17
+			GET_DEBUG_INFORMATION = 17,
+			HAS_DEBUG_INFO_LOADED = 18,
 		}
 
 		enum CmdModule {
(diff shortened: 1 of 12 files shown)


Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
```diff
@@ -566,6 +568,8 @@ enum CmdAssembly {
 			GET_TYPE_FROM_TOKEN = 11,
 			GET_METHOD_FROM_TOKEN = 12,
 			HAS_DEBUG_INFO = 13,
+			GET_CATTRS = 14,
+			GET_DEBUG_INFORMATION = 16
```
lambdageek, line 572: Why don't we have 15 here? 
https://github.com/dotnet/runtime/pull/82555

lambdageek (approved):

Mono.Debugging.Soft/SoftDebuggerBacktrace.cs
```diff
@@ -34,18 +34,60 @@
...
+						using (var response = HttpClient.GetStreamAsync (location.SourceLink.Uri).Result) {
+							Directory.CreateDirectory (Path.GetDirectoryName (sourceLinkCachedPath));
+							var fileStream = File.Create (sourceLinkCachedPath);
+							response.CopyTo (fileStream);
+							fileStream.Close ();
+							this.UpdateSourceFile (sourceLinkCachedPath);
+						}
+					}
+					catch (Exception) {
+						//expected exception when file is not available for download
```
ilonatommy, line 78: What about logging at least a message?
```suggestion
						//expected exception when file is not available for download
						DC.DebuggerLoggingService.LogMessage ($"File with uri={location.SourceLink.Uri} is not available for download);

```

ilonatommy (approved):

PR #383, opened 2023-03-02
Support receiving step from different threads

thaystg:
Use the received request and not the last sent to send to clear_request event, to avoid missing step from different threads


thaystg:
@mauroa

mauroa (approved):

PR #385, opened 2023-03-08
[Mono.Debugger.Soft] Prevent `IndexOutOfRangeException` with `MethodBodyMirror.ReadCilBody`

sailro:
Opcode cache is not big enough to handle extra reserved opcodes (`Prefixref`, `Prefix1`, ..., `Prefix7`), unused in real-world assemblies.

So prevent `IndexOutOfRangeException` by filtering those internal opcodes.


sailro:
@thaystg

thaystg:
@mauroa

jbevain:
Note: this is only used in the cased where ENABLE_CECIL is not defined. To my knowledge, ENABLE_CECIL is defined for all MDS based debuggers except VSTU.

PR #386, opened 2023-03-23
Don't let the null strings in the ProcessInfo constructor

marshall-lerner:


sandyarmstrong (approved):

Mono.Debugging/Mono.Debugging.Client/ProcessInfo.cs
```diff
@@ -64,14 +64,14 @@ internal void Attach (DebuggerSession session)
...
-			this.description = description;
+			this.name = name ?? string.Empty;
+			this.description = description ?? string.Empty;
 		}
 
 		public ProcessInfo (long id, string name)
 		{
 			this.id = id;
-			this.name = name;
+			this.name = name ?? string.Empty;
```
spouliot, line 74: this leaves the `description` field (and the associated property) to return `null`, which is what you seem to want to avoid, right ?

marshall-lerner: You're right. The `description` field wasn't the one causing issues (it was just `name`), so I'll revert the change to it

spouliot (approved):

Issue #389, opened 2023-04-06
Mono.Debugger.Soft references Microsoft.SymbolStore and Microsoft.FileFormats which don't exist on nuget.or


CartBlanche:
I was just trying to update our dependencies on `Mono.Debug*`, to the latest version on `main`. 
When I try to build `debugger-libs.sln` it tells me these 2 packages don't exist/can't be found. I checked on nuget.org and they are not there either. So `Mono.Debugger.Soft.csproj` does not compile.

```
<PackageReference Include="Microsoft.SymbolStore" Version="1.0.411401" />
<PackageReference Include="Microsoft.FileFormats" Version="1.0.411401" />
```

Any suggestions?

CartBlanche:
@thaystg You seem to have added these dependences 2 months ago (https://github.com/mono/debugger-libs/commit/4b04e4e9dc0375001965e40c953e0cbe67c4e49c). 
How do we access these new packaged if they are not published on nuget.org ?

thaystg:
Can you use from this feed?
https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json

CartBlanche:
@thaystg sure, it would just be nice if it just compiled out of the box :) Would a nuget.config help here?

PR #390, opened 2023-04-06
This will add the non-default nuget key to avoid compilation error.

thaystg:
Fixes https://github.com/mono/debugger-libs/issues/389

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,5 +1,6 @@
 <configuration>
     <config>
         <add key="repositoryPath" value="packages" />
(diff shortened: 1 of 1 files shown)


NuGet.config
```diff
@@ -1,5 +1,6 @@
 <configuration>
     <config>
         <add key="repositoryPath" value="packages" />
+        <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json">
```
MohitSethi99, line 4: I think you forgot to add `</add> in the end of this line, or perhaps a `/` before ending the tag

thaystg: Thanks!

mauroa (approved):

PR #391, opened 2023-04-06
[release-17.6-vsmac] Prevent NREs when buggy backend returns null frames

sandyarmstrong:


sandyarmstrong:
Main version of this PR is https://github.com/mono/debugger-libs/pull/392

mrward (approved):

PR #392, opened 2023-04-06
Prevent NREs when buggy backend returns null frames

sandyarmstrong:


sandyarmstrong:
VSMac 17.6 backport is https://github.com/mono/debugger-libs/pull/391

mrward (approved):

PR #412, opened 2024-08-21
[build] add extension point for consuming repositories

jonathanpeppers:
Context: https://github.com/dotnet/android/pull/9237
Context: https://github.com/dotnet/android-tools/commit/34e98e2b65917d105169f868b5648f67e68b6784

Using the latest .NET 9 RC 1 builds, dotnet/android repo fails to restore with:

    external/debugger-libs/Mono.Debugging/Mono.Debugging.csproj :
    error NU1102: Unable to find package Microsoft.NETCore.App.Ref with version (= 6.0.34

To solve this, we generally set `$(RestoreAdditionalProjectSources)` to a feed that would include the 6.0.34 runtime packages.

Add an "extension point" to mono/debugger-libs, that would allow us to have a submodule like:

    external/debugger-libs

Then would allow us to create the file:

    external/debugger-libs.override.props

This file would set `$(RestoreAdditionalProjectSources)` until the 6.0.34 runtime is released.


thaystg:
@mauroa can you please check if this will not cause any side effect on our internal extension?

PR #419, opened 2026-08-20
Rename FieldValueReference.field to avoid the C# 14 'field' keyword

csm101:
## Summary

- `Mono.Debugging.Soft/FieldValueReference.cs` has a private member named `field` that is used inside the `Name`, `Type`, `DeclaringType` and `Value` property accessors.
- Starting with C# 14 (the default language version when targeting `net10.0`), `field` is a contextual keyword inside property accessors and denotes the compiler-generated backing field, so the project fails to build with 10 `CS1061`/`CS1503` errors (e.g. `'string' does not contain a definition for 'Name'`).
- This renames the member to `fieldInfo`. Constructor and helper parameters keep their names (they are not affected). No behavior change.

Consumers that build these sources with a newer TFM currently have to pin `LangVersion` to 13; with this change they no longer need to.

## Test plan

- [x] `Mono.Debugger.Soft`, `Mono.Debugging`, `Mono.Debugging.Soft` build with `TargetFrameworks=net10.0` (C# 14) on .NET SDK 10.0.301
- [x] Still builds with the repository defaults (`net6.0;net472`)
- [x] Smoke-tested against a .NET for Android (MonoVM) app: attach, breakpoint hit, locals/fields read correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


csm101:
@dotnet-policy-service agree company="MCA Software s.a.s. di Sirna Carlo &C."

